PR #11, opened 2019-07-20
fix: set default template cte

rhnvrm:


knadh:
Two issues here:

- `= u.id` is incorrect and fails to compile. Throws `Error preparing query 'set-default-template': pq: missing FROM-clause entry for table "u"`. The correct syntax is ` = (SELECT id FROM u)`
- `id != $1` is correct and substituting it with the ID from the CTE changes the query behaviour causes the constraint to fail.

Explanation here: https://www.postgresql.org/docs/10/queries-with.html
> All the statements are executed with the same snapshot (see Chapter 13), so they cannot "see" one another's effects on the target tables.

PR #18, opened 2019-10-01
feat: read config from env

mr-karan:
Adds support to read config variables from env.

### DIFF
--- a/main.go
+++ b/main.go
@@ -15,6 +15,7 @@ import (
 	"github.com/knadh/koanf"
 	"github.com/knadh/koanf/maps"
 	"github.com/knadh/koanf/parsers/toml"
+	"github.com/knadh/koanf/providers/env"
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -112,6 +113,13 @@ func init() {
 			logger.Fatalf("error loadng config: %v.", err)
 		}
 	}
+	// Load environment variables and merge into the loaded config.
+	if err := ko.Load(env.Provider("LISTMONK_", ".", func(s string) string {
+		return strings.Replace(strings.ToLower(
+			strings.TrimPrefix(s, "LISTMONK_")), "__", ".", -1)
+	}), nil); err != nil {
+		log.Printf("error loading env config: %v", err)
```
knadh: I think this should be `log.Fatalf()`

mr-karan: Fixed.

PR #29, opened 2019-10-25
chore: minor refactors based on static checks

rhnvrm:
- unchecked returns fixed (most)
- remove unused constants
- remove unsed structs
- function parameters unused or incorrectly used
- removed if else chains for error checks
- use regex MustCompile instead of compile
- spell checks
- preallocate slice cap when size known
- scope issues inside range


rhnvrm:
Updated

Issue #79, opened 2020-04-09
Plain text alternative for campaigns


gerricom:
Hi! Do you have plans to integrate a plain text alternative for campaigns? 
Spam filters tend to be allergic to mails without a propper plain text alternative. So I'd like to see either an automatic way to add a plain text message or a method to add your own message.
Thanks!

knadh:
Hi @gerricom,
This is indeed on my to-do. I'll try to roll this out sometime soon. Prioritising items ad-hoc as requests come in :)

jaredfolkins:
PR for this feature has been submitted. It takes the HTML body, removes the HTML while keeping a similar formatting. It includes the Plain Text along with the HTML body.

Thanks,
Jared

knadh:
#80 adds this.

PR #80, opened 2020-04-09
enhc: add plain text email body along with HTML increasing deliverability

jaredfolkins:
I saw issue https://github.com/knadh/listmonk/issues/79 and as I need this functionality, I figured I'd offer a PR.

Thanks,
Jared


knadh:
Excellent! I think we should make this optional though. A `plaintext_email = true` option in [the config](https://github.com/knadh/listmonk/blob/b8d9ab10f29d8fc3c3dc0fcfb235ce453d8c58bf/config.toml.sample#L21) would be perfect for now. In the future, it can be a per-campaign option as well.

jaredfolkins:
Actually, would you consider, `email_format = {plain|html|both}`?

jaredfolkins:
>  In the future, it can be a per-campaign option as well.

Hehe. I actually started here. I have a lot of the code done but then I realized that functions like `Push()` would most likely need their signatures changed substantially.

knadh:
`email_format = {plain|html|both}` is the ideal scenario. 

`Push()` is a bit ugly. It has too many params and should've been refactored earlier. We can introduce a `messenger.Message` struct and have `Push()` just take that.

jaredfolkins:
Please review. Thanks!

knadh:
Cheers. Will push a release soon with a few other things.

PR #84, opened 2020-04-18
bug - dont update existing list

thejaswinipathi:
Suppose if a user is in mailing list A and has optin-confirmed, and now I create a new list called B, and if I add this user under B list, Previously confirmed also becomes unconfirmed.


knadh:
Good spot.

The update should still happen but instead of `DO NOTHING`, it should be:
`SET status = (CASE WHEN $4='blacklisted' THEN 'unsubscribed'::subscription_status ELSE subscriber_lists.status END);`

knadh:
@thejaswinipathi could you please update your PR?

thejaswinipathi:
https://github.com/knadh/listmonk/pull/86
Moved to a new PR, had re forked your repo, could not raise it to the same PR again..

Issue #121, opened 2020-05-31
Support for custom email headers


RaghavSood:
SMTP relay services such as AWS SES and other commercial options use custom email headers to select various relay related configurations, such as which IP address to send from, which suppression lists to enforce, etc.

Being able to specify a list of additional email header and value pairs that are added to each email would help with integrating with such services. This would work best if added as an option to the SMTP blocks in the config, since each SMTP server may have different custom headers.

knadh:
This should be straight forward to add. Will add in the next release.

RaghavSood:
Seemed like an easy enough addition so I gave it a shot, I've opened a pull request. Appreciate all your help so far!

OctavioXII:
Would this also work for the List-Unsubscribe header? This would help mitigate the bounce rate, it would be wonderful.

knadh:
@OctavioXII The above mentioned feature is for static headers. `List-Unsubscribe` should contain the dynamic unsubscribe URL. This is available in the upcoming release.

PR #122, opened 2020-05-31
Add support for custom headers

RaghavSood:
Implements #121 

Custom headers are specified as an array under `inject_headers` in config.toml. Headers are only added if both the key and value are at least 1 char, after being split by `:`.

Keys and values are not validated as the textproto package used by smtppool believes it to be better to send invalid headers rather than drop them, which also matches the behaviour of most other mail clients.

I have tested it with the `X-SES-CONFIGURATION-SET` header used by AWS to select a particular SES configuration to send the mail under, as well as sending with no additional headers, which behaves identically to the pre-PR code base. 

Signed-off-by: raghavsood <r@raghavsood.com>

### DIFF
--- a/config.toml.sample
+++ b/config.toml.sample
@@ -110,11 +110,15 @@ max_idle = 10
         wait_timeout = "5s"
 
         # The number of times a message should be retried if sending fails.
-	    max_msg_retries = 2
+	max_msg_retries = 2
 
         # Enable STARTTLS.
         tls_enabled = true
         tls_skip_verify = false
+
+        # Array of headers to add to each email sent from this server
+        # omit if you don't want to add any custom headers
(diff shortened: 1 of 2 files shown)


knadh:
I'd almost done this earlier today. Just finished up and pushed https://github.com/knadh/listmonk/commit/7a467a5a3b2adfc252dfc7f24c425e08aa4c651b. Thanks for the PR though.

Needn't to string parsing. You can define headers as maps:
`email_headers = { "X-Sender" = "listmonk", "X-Custom-Header" = "listmonk" }`

RaghavSood:
Right, that makes sense - I was trying to look for something that would translate to `map[string]string` in toml, but couldn't quite get it - no terribly familiar with toml

Thank you!

PR #151, opened 2020-07-08
frontend: pass advanced query to API when enabled

### LINKED ISSUE #150: Subscriber query does not pass query to the API
When we run an advanced query like:

```
subscribers.attribs->>'country' = 'XX'
```

The resulting api call is just `/api/subscribers?list_id=&query=&page=1` - the query isn't passed through to the API.

Expected behaviour:

The query runs successfully and the returned results are appropriately filtered, like in the old UI

RaghavSood:
Fixes #150

Apologies if this is the wrong way to do it, frontend isn't my strong point - feel free to close if there's a better option, did a quick patch for now since it was holding us up.

This also fixes a bug that caused the query to not be passed to the API when adding/removing `Select all` users from a list, which isn't mentioned in the original issue since I didn't discover it until after fixing the search.

### DIFF
--- a/frontend/src/views/Subscribers.vue
+++ b/frontend/src/views/Subscribers.vue
@@ -414,6 +414,10 @@ export default Vue.extend({
 
     // Turns the value into the simple input field into an SQL query expression.
     fullQueryExp() {
+      if (this.isSearchAdvanced) {
+        return this.queryParams.fullQuery;
+      }
+
       const q = this.queryParams.query.replace(/'/g, "''").trim();
       if (!q) {
         return '';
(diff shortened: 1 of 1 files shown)


knadh:
Your PR was correct, but I realised there scope for refactoring and simplifying a few things. https://github.com/knadh/listmonk/commit/ab8dbd8314777d00275cf683499f4aadfd457787 fixes this.

RaghavSood:
Thanks! Looks like the bugfix for passing the query when bulk/adding or removing from lists got left out, I'm reimplemented that in #155

Issue #56, opened 2020-02-06
Is it possible to embed jpg to html?


marcin-github:
Hello,
Is it possible to embed images into html?  Mail clients doesn't like images from external www servers and usually blocks it. Embedding makes email larger but it allow display mail without accepting "show remote content".

knadh:
This isn't supported right now, but I'll explore adding support. Thank you.

cb3inco:
I found a image encoder (base64) and was able to insert that into my template (https://www.base64-image.de/). Or are you looking for Listmonk to do it automatically? The site provides the img src html and css (url)

knadh:
@cb3inco it'll be a feature on listmonk  that works out of the box.

AnaghHegde:
@knadh Is this part of hactoberfest? Are you still interested in this feature ?

knadh:
Sure. If you're interested in working on this, please post a proposal on how it'll be implemented before sending a PR.

wweverma1:
@knadh is anyone currently working on this ?
If not can I start working on this feature?

fsainovich:
I can't find this in 2.2 version. Is this working ?

knadh:
This feature doesn't exist (yet) in listmonk @fsainovich.

ChrisTG742:
Is there a planned release version for this feature yet?

knadh:
This is being tracked here: https://github.com/knadh/listmonk/issues/241

PR #158, opened 2020-07-13
WIP: native support for base64 file embedding

joicemjoseph:
ref: #56
Sample output from api


knadh:
It's better to have `Get()` and `GetData()`, instead of a single `Get()` that may return a URL or encoded data.

knadh:
Closing stale PR. Feel free to reopen later when the latest version is rebased.

PR #181, opened 2020-08-24
fix: inconsistent active link on sidebar navigation

scmmishra:
## PR Summary

This PR has a fix for inconsistent active sidebar link on routing to links not in groups.

<details>
<summary>Previous Behaviour</summary>
Broken
</details>

<details>
<summary>After Fix</summary>
Fixed
</details>

## The problems
**This problem occurs when navigating from a group item to a non group item**. There are two problems to be solved here.
1. The group item is not collapsed.
1. The active state is not transferred.

### The group item not collapsed

We watch the route and change the `activeGroup` based on the current route. The `expanded` attribute of the `b-menu-item` component is overridden to toggle based on the value of the relevant key in the `activeGroup` object. 

https://github.com/knadh/listmonk/blob/0f055eacb87030b55970f4870ff3a0b22b0dcaef/frontend/src/App.vue#L135-L142

Here, on navigating to a non group item, the route `meta` does not have a `group` attribute, this means that the condition that changes the expanded state is not triggered.

We can fix this by setting the `activeGroup` to `{}`.

> An alternate solution is to fetch the `from` along with `to` from the route object and explicitly set the activeGroup to `{ [from.meta.group]: false }`, but since we are reassigning the the activeGroup always, it's not prudent to do so. Besides other problems like deep watching etc may occur.

### The active state is not transferred.
For the sidebar menu items, the code overrides the `expanded` state referring to the `activeGroup` object. However, for Buefy the expanded and active state is not explicitly linked, overriding one means the other has to be managed by the developer as well. 

The fix was simple to pass the `active` prop as well.

### DIFF
--- a/frontend/src/App.vue
+++ b/frontend/src/App.vue
@@ -32,6 +32,7 @@
                 </b-menu-item><!-- dashboard -->
 
                 <b-menu-item :expanded="activeGroup.lists"
+                  :active="activeGroup.lists"
                   icon="format-list-bulleted-square" label="Lists">
                   <b-menu-item :to="{name: 'lists'}" tag="router-link"
                     :active="activeItem.lists"
@@ -43,6 +44,7 @@
                 </b-menu-item><!-- lists -->
 
                 <b-menu-item :expanded="activeGroup.subscribers"
+                  :active="activeGroup.subscribers"
                   icon="account-multiple" label="Subscribers">
                   <b-menu-item :to="{name: 'subscribers'}" tag="router-link"
                     :active="activeItem.subscribers"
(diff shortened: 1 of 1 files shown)


frontend/src/App.vue
```diff
@@ -54,6 +56,7 @@
                 </b-menu-item><!-- subscribers -->
 
                 <b-menu-item :expanded="activeGroup.campaigns"
+                  :active="activeGroup.campaigns"
                     icon="rocket-launch-outline" label="Campaigns">
```
knadh, line 60: While you are at it, could you fix the indentation here?

knadh:
This PR is so well documented, I shed tears of joy :sob:. Thanks!

PR #189, opened 2020-09-16
refactor: switch to internal app package from main

rhnvrm:
This commit moves all the logic in the main package
to a new internal app package and adds a new simpler main.go
to the cmd/app main package.

With this refactor we remove the global state during the init and
scope the global vars that we had earlier. This might help with
testing (if needed) later on as well.

The biggest benefit of this refactor for new users and developers
is running `go run cmd/app/main.go` to get started with listmonk.

----

#

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,15 +20,15 @@ deps:
 # Build steps.
 .PHONY: build
 build:
-	go build -o ${BIN} -ldflags="-s -w -X 'main.buildString=${BUILDSTR}' -X 'main.versionString=${VERSION}'" cmd/*.go
+	go build -o ${BIN} -ldflags="-s -w -X 'main.buildString=${BUILDSTR}' -X 'main.versionString=${VERSION}'" ./cmd/listmonk
 
 .PHONY: build-frontend
 build-frontend:
 	export VUE_APP_VERSION="${VERSION}" && cd frontend && yarn build
 
 .PHONY: run
-run: build
-	./${BIN}
+run: 
+	go run ./cmd/listmonk	
 
 .PHONY: run-frontend
 run-frontend:
--- a/internal/app/app.go
+++ b/internal/app/app.go
@@ -1,4 +1,4 @@
-package main
+package app
 
 import (
 	"context"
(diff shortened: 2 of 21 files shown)


joicemjoseph:
is this inspired by Matt's post - [why you shouldnt use func main](https://pace.dev/blog/2020/02/12/why-you-shouldnt-use-func-main-in-golang-by-mat-ryer.html)

rhnvrm:
> Asking out of curiosity, is this inspired by Matt's post - [why you shouldnt use func main](https://pace.dev/blog/2020/02/12/why-you-shouldnt-use-func-main-in-golang-by-mat-ryer.html)

Just read the article and it seems to have some good suggestions, although it seems to be having different goals. My aim with this PR was simple, to refactor and remove the globals and just allow me as a developer to run `go run ./cmd/listmonk` and `go build ./cmd/listmonk` and keep the `cmd/` folder conventional.

knadh:
Makes sense! I'll review this over the weekend.

Makefile
```diff
@@ -20,15 +20,15 @@ deps:
 # Build steps.
 .PHONY: build
 build:
-	go build -o ${BIN} -ldflags="-s -w -X 'main.buildString=${BUILDSTR}' -X 'main.versionString=${VERSION}'" cmd/*.go
+	go build -ldflags="-s -w -X 'main.buildString=${BUILDSTR}' -X 'main.versionString=${VERSION}'" ./cmd/listmonk
```
knadh, line 23: Let's use the `$BIN` variable here to be explicit (`-o ${BIN}`). The `make dist` also uses this.

internal/app/app.go
```diff
@@ -48,21 +48,11 @@ type App struct {
 	sync.Mutex
 }
 
-var (
-	lo = log.New(os.Stdout, "", log.Ldate|log.Ltime|log.Lshortfile)
+func initApp(buildString, versionString string) (fs stuffbin.FileSystem, db *sqlx.DB, ko *koanf.Koanf, lo *log.Logger) {
```
knadh, line 51: We should replace the naked return here with an explicit return. Confusing otherwise.

rhnvrm:
@knadh 
I have resolved the suggested changes.

knadh:
Cheers. Let's wait for the `postbacks` branch to be merged into master, on top which, this refactor can be applied.

knadh:
This branch has diverged too much. Closing this PR. After the next release, we can discuss and do this refactor before further changes.

PR #205, opened 2020-10-11
fix: use mail.ParseAddress to validate email instead of custom regex

vividvilla:
According to [RFC6532](https://tools.ietf.org/html/rfc6532) unicode characters are supported in email address but the in-built custom regex was only validating ASCII characters.

### DIFF
--- a/internal/subimporter/importer.go
+++ b/internal/subimporter/importer.go
@@ -17,6 +17,7 @@ import (
 	"io"
 	"io/ioutil"
 	"log"
+	"net/mail"
 	"os"
 	"regexp"
 	"strings"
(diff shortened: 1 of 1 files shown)


internal/subimporter/importer.go
```diff
@@ -619,5 +617,8 @@ func countLines(r io.Reader) (int, error) {
 
 // IsEmail checks whether the given string is a valid e-mail address.
 func IsEmail(email string) bool {
-	return regexEmail.MatchString(email)
+	if _, err := mail.ParseAddress(email); err != nil {
```
knadh, line 620: `mail.ParseAddress()` parses an "address", not an e-mail. For instance `Name <name@email.com>`, `<name@email.com>` and `name@email.com`. However, in this context, we're validating just the e-mail ID component. 

Adding an `email == address.Email` check to the parsed address should address this.

vividvilla: Actully we can just check if `email.Name` is not empty to validate if its not the address. For example

```
func IsEmail(email string) bool {
	if em, err := mail.ParseAddress(email); err != nil || em.Name != "" {
              return false
	}
	return true
}
```

knadh: Doesn't work with `<bécassine@example.com>` though. Angle brackets with no name.

knadh: Sorry, missed this. It should be `len(addrs)` and not `len(addrs)-1`, right?

vividvilla: Yeah, you are right. Doing `email == address.Address` is the better way to do this - https://play.golang.org/p/drvQgi6JdW1

will update PR to that?

knadh: Yep.

vividvilla: Updated the PR.

Issue #225, opened 2020-11-17
bug: DB Connections not being reused


mr-karan:
## Bug

In the sample config files provided for [binary](https://github.com/knadh/listmonk/blob/master/config.toml.sample) and [docker](https://github.com/knadh/listmonk/blob/master/config-demo.toml) installations, there's no configuration set for `db.max_open` and `db.max_idle` parameters.

Since these values are not set, they are by default passed as `0` when unmarshalled and configured [here](https://github.com/knadh/listmonk/blob/1aecd6f2e13d7a07c4938e280723430e1bc1e170/cmd/queries.go#L105).

While having an unbounded connection is not desirable, it's not _really_ that problematic but setting `db.SetMaxIdleConns(0)` means that every connection is established from scratch, thus not utilizing the connection pooling at all.

Here's the `pg_stat_activity` query results:

```
SELECT
  COUNT(*),
  state
FROM pg_stat_activity
WHERE datname = 'listmonk'
GROUP BY 2;
```

- When `max_idle` isn't specified (0):

```
 count | state  
-------+--------
     1 | active
```

- When `max_idle` is specified (25):

```
 count | state  
-------+--------
     1 | active
     2 | idle
```

## Proposed Fix

I think we can provide the following options as "sane defaults" in the sample configs.

```
max_open 25
max_idle 25
```

## Unrelated

In addition to this, and although this is unrelated and not a bug, but we should also set `db.SetConnMaxLifetime` to something like 5-10 minutes.

knadh:
Makes sense. Can you send a PR? We just need this added to the sample config + db init. Default for older config after the next upgrade will be 0 automatically, so nothing changes.

PR #226, opened 2020-11-17
Fix/db config

mr-karan:
Closes https://github.com/knadh/listmonk/issues/225

### DIFF
--- a/cmd/queries.go
+++ b/cmd/queries.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
+	"time"
 
 	"github.com/jmoiron/sqlx"
 	"github.com/lib/pq"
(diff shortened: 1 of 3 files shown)


config-demo.toml
```diff
@@ -10,3 +10,6 @@
     password = "listmonk"
     database = "listmonk"
     ssl_mode = "disable"
+    max_open = 25
+    max_idle = 25
+    max_conn_lifetime = 300 # 5min or 300s.
```
knadh, line 15: A timestring here is better. eg: `max_conn_lifetime = "300s"`

cmd/queries.go
```diff
@@ -84,14 +85,15 @@ type Queries struct {
...
-	MaxIdle  int    `koanf:"max_idle"`
+	Host            string        `koanf:"host"`
+	Port            int           `koanf:"port"`
+	User            string        `koanf:"user"`
+	Password        string        `koanf:"password"`
+	DBName          string        `koanf:"database"`
+	SSLMode         string        `koanf:"ssl_mode"`
+	MaxOpen         int           `koanf:"max_open"`
+	MaxIdle         int           `koanf:"max_idle"`
+	MaxConnLifetime time.Duration `koanf:"max_conn_lifetime"`
```
knadh, line 96: Maybe `max_lifetime` to be consistent with `max_open` and `max_idle`?

config-demo.toml
mr-karan: Yep. Just realised it was setting it as `300ns` otherwise. Phew. :sweat_smile: 

Fixed.

PR #262, opened 2021-02-01
Add Portuguese Translations

diogotcorreia:
This PR adds Portuguese (European) translations.

It also fixes some typos in the English file.


enniosousa:
@diogotcorreia great! 

I would like add Brazilian Portuguese too.

I think that use a languages standart code is recomended, then allow dialects of the base language from a particular location.

> The full [BCP47](https://www.ietf.org/rfc/bcp/bcp47.txt) syntax is in-depth enough to mark extremely specific language dialects, but most usage is much simpler.
>
>A language tag is made of hyphen-separated language subtags, where each subtag indicates a certain property of the language. The 3 most common subtags are:
>
>Language subtag
>Required. A 2-or-3-character code that defines the basic language, typically written in all lowercase. For example, the language code for English is en, and the code for Badeshi is bdz.
>
>Script subtag
>Optional. This subtag defines the writing system used for the language, and is always 4 characters long, with the first letter capitalized. For example, French-in-Braille is fr-Brai and ja-Kana is Japanese written with the Katakana alphabet. If the language is written in a highly typical way, like English in the Latin alphabet, there is no need to use this subtag.
>
>Region subtag
>Optional. This subtag defines a dialect of the base language from a particular location, and is either 2 letters in ALLCAPS matching a country code, or 3 numbers matching a non-country area. For example, es-ES is for Spanish as spoken in Spain, and es-013 is Spanish as spoken in Central America. “International Spanish” would just be es.
The script subtag precedes the region subtag if both are present — ru-Cyrl-BY is Russian, written in the Cyrillic alphabet, as spoken in Belarus.
>
>To find the correct subtag codes for a language, try the Language Subtag Lookup.
> Ref.: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang

diogotcorreia:
@enniosousa I've seen that in some projects there is `pt` and `pt-br`, for example.
I've used just `pt` since the developer used `en` for English (without specifying US/UK) and the same for other languages.

If it's something the developer would like me to change, I'd be fine with it.

knadh:
Great @diogotcorreia!

@enniosousa the two letter codes are ISO 639-1. Did some digging and `pt-BR` is appropriate here (which also happens to be the BCP47) code.

PR #306, opened 2021-03-30
Fix: campaign field names

kousikmitra:
Issue:

image

1. Added : for placeholder property
2. Added `rawHTML` under `fields` namespace
https://github.com/knadh/listmonk/blob/6dbcfee0801148c23cbb0444d15c7dce28bd672d/frontend/src/views/TemplateForm.vue#L19

### DIFF
--- a/frontend/src/views/TemplateForm.vue
+++ b/frontend/src/views/TemplateForm.vue
@@ -13,7 +13,7 @@
         <section expanded class="modal-card-body">
             <b-field :label="$t('globals.fields.name')" label-position="on-border">
             <b-input :maxlength="200" :ref="'focus'" v-model="form.name"
-                placeholder="$t('globals.fields.name')" required></b-input>
+                :placeholder="$t('globals.fields.name')" required></b-input>
             </b-field>
 
             <b-field :label="$t('globals.fields.rawHTML')" label-position="on-border">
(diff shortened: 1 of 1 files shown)


knadh:
Thanks @kousikmitra. The `rawHTML` field should not be in globals though, but `campaigns`, as it's used in just a couple places. I think this is already fixed in the [testing branch](https://github.com/knadh/listmonk/tree/testing). Could you compare with that and amend?

kousikmitra:
> Thanks @kousikmitra. The `rawHTML` field should not be in globals though, but `campaigns`, as it's used in just a couple places. I think this is already fixed in the [testing branch](https://github.com/knadh/listmonk/tree/testing). Could you compare with that and amend?

Changed the base to testing, its not changed yet. Will change it to `globals.campaigns.rawHTML`

knadh:
It shouldn't be `globals` at all. It's already present here: https://github.com/knadh/listmonk/blob/c90d42a7a9fa2d08a5190f55dea895530bdfd5a1/i18n/en.json#L45

https://github.com/knadh/listmonk/blob/c90d42a7a9fa2d08a5190f55dea895530bdfd5a1/i18n/en.json#L430

kousikmitra:
fixed it.

PR #309, opened 2021-04-10
Bundle sprig template functions

tamalsaha:
Signed-off-by: Tamal Saha <tamal@appscode.com>


tamalsaha:
I think it will useful to have these functions available to generate templates. https://github.com/Masterminds/sprig/blob/master/docs/index.md

knadh:
Didn't know about sprig. This is very useful. Adding the map on every compilation allocates a bunch and adds an average ~150μs just to the template func construction (got curious and checked), but that's negligible compared to the actual I/O bottleneck of messages. Merging this. Thanks.

PR #310, opened 2021-04-11
Enable auto TLS using certificate issued from let's encrypt

tamalsaha:
ref: https://echo.labstack.com/cookbook/auto-tls/

Signed-off-by: Tamal Saha <tamal@appscode.com>

### DIFF
--- a/config.toml.sample
+++ b/config.toml.sample
@@ -1,6 +1,6 @@
 [app]
     # Interface and port where the app will run its webserver.
-    address = "0.0.0.0:9000"
+    address = "0.0.0.0:443"
 
     # BasicAuth authentication for the admin dashboard. This will eventually
     # be replaced with a better multi-user, role-based authentication system.
@@ -8,6 +8,12 @@
     # only where an external authentication is already setup.
     admin_username = "listmonk"
     admin_password = "listmonk"
+
+[tls]
+    enable = true
+    domain = "listmonk.company.com"
+    cert_dir = "/app"
+    email = "me@company.com"
 
 # Database.
 [db]
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"html/template"
+	"net/http"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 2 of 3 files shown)


knadh:
Nice. Since this involves new config keys, the settings should be added to the UI. If you don't want to patch the frontend, I can merge this and add the UI changes.

tamalsaha:
That sounds good. Please feel free to make changes to the UI.

config.toml.sample
```diff
@@ -9,6 +9,12 @@
     admin_username = "listmonk"
     admin_password = "listmonk"
 
+[tls]
+    enable = true
+    domain = "listmonk.company.com"
```
alerque, line 14: This would be duplicate info that is already entered in the UI and saved in the database, but I don't think this setting should go here at all or where it is either. See #346 for why.

config.toml.sample
```diff
@@ -1,6 +1,6 @@
 [app]
     # Interface and port where the app will run its webserver.
-    address = "0.0.0.0:9000"
+    address = "0.0.0.0:443"
```
alerque, line 3: It should not be assumed or demonstrated by default that listmonk runs on 80/443. The current example using a non privileged port by default is better. Otherwise you need to run as root, and that should not be suggested by default. I suggest reverting this back, and just document in a comment nearby the need to run on 80/443 if using the built in SSL cert features.

Also note that if using the TLS feature not one but 2 ports with be used by the app, and that should be reflected in the config somehow.

config.toml.sample
```diff
@@ -9,6 +9,12 @@
     admin_username = "listmonk"
     admin_password = "listmonk"
 
+[tls]
+    enable = true
```
alerque, line 13: Please default this to "off" in the sample config. This feature should be explicitly requested by the admin, not assumed. It is too intrusive and can mess up other domain certificate arrangements.

alerque (changes requested):

knadh:
I have second thoughts about bundling LetsEncrypt support into the app as it feels out of scope.

TLS is almost always managed and provisioned externally in production environments. listmonk is also highly likely to be run behind a proxy like Nginx, Apache, or Caddy, where certbot can be trivially setup.

alerque:
I completely concur and would prefer *not* to see this support bundled. My responses here were to make sure the feature set doesn't bungle things up for what I agree is the majority of production use cases where actual SSL termination for privileged ports on domains is not going to be handled by this daemon directly.

This could perhaps be something bundled in to an alternate Docker container for a "top level deploy ready" scenario, but I would be more comfortable with it being completely separate from the app itself.

tamalsaha:
Feel free to close the pr if you don't intend to merge it.

knadh:
@tamalsaha thanks again for the PR, but after giving it some thought, I think it's best if this isn't integrated into the core. Closing this.

PR #317, opened 2021-04-14
French translations

FelixDz:
Added / corrected french translations.


FelixDz:
Hmm, not sure about the branch I'm merging this into, considering the gap with master. Should it be master instead then ?

knadh:
Thanks, the PR looks fine. I only remember about 10 words from the two years worth of French classes I took in highschool, so can't comment on the language! :smile: 

`master` is the correct branch.

PR #341, opened 2021-05-05
Prevent mistakes adding the OS and Arch on build

brunowego:
After minutes trying understand why my build are not working, I found where I'm doing wrong, so, I build on Darwin and run the binary on Linux. Just to prevent this kinda of mistakes, I added the flags `GOOS` and `GOARCH` explicit on the Makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ deps:
 # Build the backend to ./listmonk.
 .PHONY: build
 build:
-	CGO_ENABLED=0 go build -o ${BIN} -ldflags="-s -w -X 'main.buildString=${BUILDSTR}' -X 'main.versionString=${VERSION}'" cmd/*.go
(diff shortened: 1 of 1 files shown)


mr-karan:
Hey. I don't think that's a correct patch.

This makes it impossible for people on Darwin to build and run locally, should they.

brunowego:
I understand @mr-karan, perhaps a different way, more explicit way telling about the currently OS/ARCH.

alerque (commented):
his should note be hard coded like this, the current arrangement building for the current env or however the host system's Go is configured should be sufficient.

alerque:
> I understand @mr-karan, perhaps a different way, more explicit way telling about the currently OS/ARCH.

What is the problem you are trying to fix? If you have a need to be more explicit you can just export this yourself before building:

    export GOOS=linux GOARCH=amd64

... or asa you build

	GOOS=linux GOARCH=amd64 make

..but what is the problem with assuming the default / current host environment as detected by Go is correct? It seems like you might be trying to solve a problem that is out of scope for a downstream project.

brunowego:
Good explanation @alerque, ok, now understand. Thanks :)

PR #349, opened 2021-05-07
Pass through raw inline HTML from Markdown content

alerque:
Should be "safe" in the sense that we are not parsing untrusted input
here. Any user that can enter Markdown can also switch to raw HTML input
mode anyway, all this does is allow the Markdown editor to be used for
more use cases without having to switch to a different input format.


knadh:
Will test and merge this over the weekend.

PR #350, opened 2021-05-07
Cleanup sample config

alerque:
See commit message comments for details. The TOML formatting wasn't invalid, it's just non-traditional and annoyed me a lot. The other two changes to sample configs are just sysadmin convenience and more safety out of the box.

### DIFF
--- a/config.toml.sample
+++ b/config.toml.sample
@@ -1,6 +1,6 @@
 [app]
 # Interface and port where the app will run its webserver.
-address = "0.0.0.0:9000"
+address = "localhost:9000"
 
 # BasicAuth authentication for the admin dashboard. This will eventually
 # be replaced with a better multi-user, role-based authentication system.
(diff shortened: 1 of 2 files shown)


knadh:
This is a vestige of the indentation I applied when there were dozens and dozens of keys and it was painful to read and parse mentally. Thankfully, all of that has now moved to the UI.

config.toml.sample
```diff
@@ -1,22 +1,22 @@
 [app]
-    # Interface and port where the app will run its webserver.
-    address = "0.0.0.0:9000"
+# Interface and port where the app will run its webserver.
+address = "localhost:9000"
```
knadh, line 3: Should explicitly document `0.0.0.0`. Something along the lines of, `To make listmonk accessible publicly over the internet, bind to 0.0.0.0:9000` or something. A vast majority of users who deploy are neither sysadmins nor developers, just hobbyists and this may not be apparent to them.

alerque: True, it should be documented how to make the port accessible from the main IP, but I still think defaulting to `localhost` and documenting how they can switch to `0.0.0.0` for all interfaces or a specific IP would be better.

Issue #368, opened 2021-05-21
SQLite Support


maxhawkins:
Any plans to support SQLite?

I don't currently run Postgres on my tiny VPS and would prefer to use a lightweight database like SQLite.

I saw in a previous issue you don't plan to support alternative databases while the app is in beta. Now that you've hit 1.0 are there any plans to change that?

SQLite might be a good choice because it will be easier for people to get up and running with the standalone binary even if they don't have a database server handy.

knadh:
No plans right now, @maxhawkins. 90% of the [SQL queries](https://github.com/knadh/listmonk/blob/master/queries.sql) would be easy to port, but the rest would be extremely hard. And, listmonk relies heavily on these queries for executing app logic.

That said, I've a personal 2 vCPU VPS instance on which I run Postgres that handles hundreds of queries per second on tables with a few million rows, serving user traffic alongside a bunch of other applications. It runs fine. In addition, listmonk is very light on resources including DB usage, so you should be able to get away with it. Of course, cannot compare to SQLite!

maxhawkins:
Yeah, Postgres / using the API probably makes the most sense. It's nice having your database as a single file but running another server isn't too bad. My other applications run on SQLite so I was hoping I could do joins across the mailing list and another application's data store but I understand wanting to only support a single dialect.

I did take a look at the queries earlier today to see if it would be easy to implement a SQLite driver myself. At first glance, with the exception of the list types, it looked like everything would translate easily. It may be possible to emulate the missing list types with SQLite's json support.

Where I ran into trouble was with the Postgres-specific pq.*Array types passed as parameters elsewhere in the code. It seemed like fixing that would add a lot of extra abstraction so I stopped. I can post the SQLite-compatible translation of the schema I made if that would be helpful at all.

knadh:
> I can post the SQLite-compatible translation of the schema I made if that would be helpful at all.

That would indeed be useful.

> At first glance, with the exception of the list types, it looked like everything would translate easily.

List and JSON types are used heavily, and there are other things like the `UNNEST()`, `COUNT(*) OVER()` which SQLite doesn't seem to support.

To support multiple DB backends, all query execution will have to be abstracted into functions on an interface. Some logic might end up moving into app code too. This is not necessarily bad, but some important considerations off the top of my head:

- The added effort of maintaining and testing multiple DB backends for all DB related features and changes may be significant. This will be non-trivial upfront and ongoing work.
[...]

PR #372, opened 2021-05-26
[WIP] Refactor database methods behind interface

maxhawkins:
This commit works toward multiple database implementations (#368) by putting all database methods inside of a Go interface.

This is still a work in progress. I am aiming to do the refactor without introducing regressions in the existing codebase by slowly changing the existing implementation and testing after refactoring each method.

### DIFF
--- a/cmd/admin.go
+++ b/cmd/admin.go
@@ -62,7 +62,7 @@ func handleGetDashboardCharts(c echo.Context) error {
 		out types.JSONText
 	)
 
-	if err := app.queries.GetDashboardCharts(&out); err != nil {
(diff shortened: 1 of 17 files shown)


maxhawkins:
@knadh I now have the Postgres methods behind a Go interface that matches the calling convention of the original SQL statements. I want to make sure the code works before cleaning up the interface so it doesn't use positional arguments.

Could you take a look? Since there aren't tests it's hard for me to know where each of the statements gets used.

knadh:
Archiving this. Will try to pick this up in the near future hopefully!

PR #374, opened 2021-05-27
Update de.json

jonathanmmm:
I have tried changing them as best as I could.
I found that in english there is about AWS written something about access key and secret key but
AWS tells on their page about  `Access keys (access key ID and secret access key) `
https://docs.aws.amazon.com/general/latest/gr/aws-sec-cred-types.html
Should the access key mean ID and the second secret access key?


knadh:
`Access key` and `Secret key` are used widely and are fine.

Issue #380, opened 2021-06-02
Health Check returns 401 when authentication is enabled


Thanzex:
**Version:**
 - listmonk: v1.0.0

**Description of the bug and steps to reproduce:**
When listmonk's basic auth is enabled, the health check endpoint is behind the authentication middleware, this returns 401 failing liveness/readiness probes.

The `/api/health` endpoint should be exempt from authentication checks

Since other public facing links, such as subscription management from the end user, are not affected this should be an easy fix

https://github.com/knadh/listmonk/blob/59bcc8eb133e8e3f7d440381a111f72c3f2ec819/cmd/handlers.go#L44-L52

PR #381, opened 2021-06-02
fix: expose healthcheck API as public endpoint

mr-karan:
Fixes https://github.com/knadh/listmonk/issues/380

If basic auth is enabled, the endpoint will still respond with 200.

```sh
curl -i localhost:9000/api/health
HTTP/1.1 200 OK
Content-Type: application/json; charset=UTF-8
Date: Wed, 02 Jun 2021 14:38:47 GMT
Content-Length: 14

{"data":true}
```

### DIFF
--- a/cmd/handlers.go
+++ b/cmd/handlers.go
@@ -151,6 +151,8 @@ func registerHTTPHandlers(e *echo.Echo, app *App) {
 		"campUUID", "subUUID"))
 	e.GET("/campaign/:campUUID/:subUUID/px.png", validateUUID(handleRegisterCampaignView,
 		"campUUID", "subUUID"))
+	// Public health API endpoint.
+	e.GET("/health", handleHealthCheck)
 }
 
 // handleIndex is the root handler that renders the Javascript frontend.


knadh:
hm, this would make `/api/health` an exception to the `/api/*` path which is meant to be private, complicating writing firewall and proxy rules.

How about we retain the old `/api/health` for backwards compatibility (and deprecate it) and add a new public handler, `/health`?

mr-karan:
> path which is meant to be private, complicating writing firewall and proxy rules.

Makes sense.

> How about we retain the old /api/health for backwards compatibility (and deprecate it) and add a new public handler, /health?

Made the changes.

```
❯ curl -i localhost:9999/health localhost:9999/api/health
HTTP/1.1 200 OK
Content-Type: application/json; charset=UTF-8
Date: Thu, 03 Jun 2021 05:53:27 GMT
Content-Length: 14

{"data":true}
HTTP/1.1 401 Unauthorized
Content-Type: application/json; charset=UTF-8
Www-Authenticate: basic realm=Restricted
Date: Thu, 03 Jun 2021 05:53:27 GMT
Content-Length: 27

{"message":"Unauthorized"}
```

PR #403, opened 2021-06-26
streamline make configuration

kmohrf:
Hi,

I’m glad I’ve found this "little" gem and I’ve been working on [a deb package](https://git.hack-hro.de/kmohrf/listmonk-deb) for the past three hours and have encountered a few bumps along the road. It comes down to these two issues:

* Debian installs yarn as `yarnpkg`. This first commit fixes this by simply allowing the caller to override the yarn binary name. This also make it easier to pass options (i.e. for npm caching servers).
* I’ve noticed that the `dist` make target will execute builds each time it’s called and is missing a dependency on the `deps` target. I’ve split up the latter into dedicated targets for `stuffbin` and `node_modules` so that each of them is only executed when necessary. I’ve also defined dependencies for the build targets so that rebuilds should only get triggered if the actual source files changed. I’m not a make-guru, so there’s probably some better way of defining these things, but I’d say it works as a first step in that direction.

Apart from the `pack-bin` and `dist` targets seem to do the same. I’ve kept the `pack-bin` target, but maybe it can be removed?

Feel free to comment on, merge, cherry-pick or gut these changes and thank you for your time!

Cheers,

Konrad

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,8 @@ LAST_COMMIT := $(shell git rev-parse --short HEAD)
 VERSION := $(shell git describe --tags --abbrev=0)
 BUILDSTR := ${VERSION} (\#${LAST_COMMIT} $(shell date -u +"%Y-%m-%dT%H:%M:%S%z"))
 
+YARN ?= yarn
+
 BIN := listmonk
 STATIC := config.toml.sample \
 	schema.sql queries.sql \
@@ -15,7 +17,7 @@ STATIC := config.toml.sample \
 .PHONY: deps
 deps:
 	go get -u github.com/knadh/stuffbin/...
-	cd frontend && yarn install
+	cd frontend && $(YARN) install
 
 # Build the backend to ./listmonk.
 .PHONY: build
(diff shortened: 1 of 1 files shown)


knadh:
This looks great! All make commands work fine on my Fedora machine.

@alerque would you be able to give this a spin in your Arch environment and see if your builds work as before?

> Apart from the pack-bin and dist targets seem to do the same. I’ve kept the pack-bin target, but maybe it can be removed?

`pack-bin` is used by goreleaser to pack the multiple cross platform binaries it produces outside of the make steps: https://github.com/knadh/listmonk/blob/master/.goreleaser.yml#L26

alerque:
A very quick read-through suggests these are definitely improvements in a useful direction. I can tell you it's not going to break my Arch builds because I had to completely eschew the `Makefile` (which, no offense intended, was terrible) and just copied and adapted the few bits of commands needed. These changes look like they might bring it to the point of being *maybe* usable with just a little more tweaking, but I'll have to play around with it for real later. Thus far I'm just commenting over a quick visual scan.

knadh:
Cool @alerque. Just wanted confirmation that these changes wouldn't break the AUR package. Merging this.

Issue #411, opened 2021-06-29
Bundle sysytem unit file in the repo


knadh:
> I’d be nice to have a default systemd unit file as part of the project that can serve as a basis for package maintainers. systemd provides a lot of options that reduce the attack surface of an application and it would be nice if we could document any practical and applicable options right here, so they’re discoverable for everyone.

Ref https://github.com/knadh/listmonk/issues/409

knadh:
@kmohrf please feel free to send a PR.

kmohrf:
Can you shed some light on how the paths are constructed? The default value for uploads in the database is `uploads`. Is that a relative directory? If so: to which directory is it relative to?

One of the most useful options for hardening a systemd service would be to use the `DynamicUser` directive. In that case systemd will create a user on-the-fly when starting the service and assigns a state/home directory to it. This state directory is kind of predictable but i guess it would help if the upload path field would either support the `$HOME` env variable. In case directories are relative to the users home it should work out-of-the-box.

knadh:
By default, the path is `./uploads` and it is relative to the binary's CWD, whichever path the bin is run from. This is just a placeholder and users are expected to create a directory and update the path in the settings UI. listmonk by default does not create an `uploads` dir, so without explicit user action, uploads don't work.

> This state directory is kind of predictable but i guess it would help if the upload path field would either support the $HOME env variable. 

If this works `https://golang.org/pkg/os/#UserHomeDir`, we can consider adding support for a home `$HOME` placeholder in all path fields on the UI. Would you be able to check if this function returns the correct path when run as a `DynamicUser` in systemd?

kmohrf:
> By default, the path is ./uploads and it is relative to the binary's CWD, whichever path the bin is run from.

Maybe I’m just confused but that seems contradictory? I install listmonk to `/usr/share/listmonk/listmonk` the binary itself and the directory are owned by root and are not writable by listmonk. I start listmonk from `/var/lib/private/listmonk-foo` (my CWD) with `/usr/share/listmonk/listmonk --config /etc/listmonk/foo.toml`. Would the default upload path refer to `/usr/share/listmonk/uploads` or `/var/lib/private/listmonk-foo/uploads`?

> If this works https://golang.org/pkg/os/#UserHomeDir, we can consider adding support for a home $HOME placeholder in all path fields on the UI. Would you be able to check if this function returns the correct path when run as a DynamicUser in systemd?

I guess so. I’ll give it a try!

[...]

Issue #409, opened 2021-06-29
packaging needs


kmohrf:
**Is your feature request related to a problem? Please describe.**

There are a few issues I’ve bumped into while packaging listmonk for debian(-based) systems. I adressed some of those [in this PR](https://github.com/knadh/listmonk/pull/403). Maybe we can use this issue to explore some of the remaining stumbling blocks.

**Describe the solution you'd like**

* The `LAST_COMMIT` and `VERSION` make variables are currently hardcoded to use git for determining their respective values. Package builds often use release tarballs which do not contain any git metadata. The easiest solution is to make the caller determine these values by using the `?=` instead of the `:=` assignment in `Makefile`. This is what I do in [my deb packaging](https://git.hack-hro.de/kmohrf/listmonk-deb/-/blob/main/Makefile#L30). The downside is that every packager has to do that manually. A lot of other projects store release metadata in the repository.
* `--new-config` should respect the `--config` option. Pratically speaking: calling `listmonk --config /etc/listmonk/mysite.toml --new-config` should create the config in `/etc/listmonk/mysite.toml` instead of storing it as `config.toml` in the current working directory.
* I know that #326 and #344 already pointed this out, but I want to emphasize it again: I think it would be nice if update checks could be disabled either via the application configuration file or – even better – during build time. It is very common for packagers to disable these checks. There shouldn’t be a need to even render the update-check switch in the frontend in both of these cases.
* I’d be nice to have a default systemd unit file as part of the project that can serve as a basis for package maintainers. systemd provides a lot of options that reduce the attack surface of an application and it would be nice if we could document any practical and applicable options right here, so they’re discoverable for everyone.

As @alerque seems to be packaging listmonk for ArchLinux he can probably weigh in on this discussion as well. Consider it a friendly request for comments :).

I’d be happy to submit pull requests if we reach consensus on solutions for any of the raised issues. Thank you for your time!

Cheers,

Konrad

knadh:
> The LAST_COMMIT and VERSION make variables are currently hardcoded to use git for determining their respective values. Package builds often use release tarballs which do not contain any git metadata. The easiest solution is to make the caller determine these values by using the ?= instead of the := assignment in Makefile. This is what I do in my deb packaging. The downside is that every packager has to do that manually. A lot of other projects store release metadata in the repository.

[...]

PR #417, opened 2021-07-07
add systemd service unit

kmohrf:
As suggested in #409 and documented in #411 this PR adds a systemd service unit as a basis for package maintainers and should even be usable as a drop-in unit for linux distributions that use systemd as its init system as long as the listmonk binary resides in `/usr/bin/listmonk`.

Any comments or questions are appreciated :).

### DIFF
--- a/listmonk@.service
+++ b/listmonk@.service
@@ -1,0 +1,69 @@
+[Unit]
+Description=listmonk mailing list and newsletter manager (%I)
+ConditionPathExists=/etc/listmonk/%i.toml
+Wants=network.target
+# The PostgreSQL database may not be on the same host but if it
+# is listmonk should wait for it to start up.
+After=postgresql.service
+
+[Service]
+Type=simple
+EnvironmentFile=-/etc/default/listmonk
+EnvironmentFile=-/etc/default/listmonk-%i
+ExecStartPre=/usr/bin/mkdir -p "${HOME}/uploads"
+ExecStartPre=/usr/bin/listmonk --config /etc/listmonk/%i.toml --upgrade --yes
+ExecStart=/usr/bin/listmonk --config /etc/listmonk/%i.toml $SYSTEMD_LISTMONK_ARGS
+Restart=on-failure
+
+# Create dynamic users for listmonk service instances
+# but create a state directory for uploads in /var/lib/private/%i.
+DynamicUser=True
+StateDirectory=listmonk-%i
+Environment=HOME=%S/listmonk-%i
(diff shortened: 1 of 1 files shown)


listmonk@.service
```diff
@@ -0,0 +1,69 @@
...
+# The PostgreSQL database may not be on the same host but if it
+# is listmonk should wait for it to start up.
+After=postgresql.service
+
+[Service]
+Type=simple
+EnvironmentFile=-/etc/default/listmonk
+EnvironmentFile=-/etc/default/listmonk-%i
+ExecStartPre=/usr/bin/mkdir -p "${HOME}/uploads"
+ExecStartPre=/usr/bin/listmonk --config /etc/listmonk/%i.toml --upgrade --yes
```
knadh, line 14: DB upgrade, while ideally safe an idempotent, is still a "destructive" action that should require explicit user consent, given the end user an opportunity to backup the DB. Not sure if it should implicitly executed.

kmohrf: Personally I’d value the continuous operation of the service higher than the little feel of control that comes with the manual execution of database migrations. Either I’m performing automated backups, in which case I would want the service to automatically execute migrations because I can always rollback or I did not configure backups and I’m taking the lets-hope-for-the-best road anyway :). But I understand that this is a debatable opinion :).

I see a few options:
1. no automatic migrations at all: In that case it would be nice if listmonk could go into a failed state if the expected migration version does not match the database version. The question is if this failed state should be visible to the system administrator or the listmonk administrator. I vote it should be the system administrator because they’re the only one who can fix it, which means that the service should exit with a non-zero exit code if migrations haven’t been executed.
2. backup before migrate: listmonk automatically performs a backup with the configured credentials before executing migrations. Possibly enabled through a cli option like `--backup`.
3. configuration-level consent: only automatically execute migrations if consent has been given through configuration (either an env variable, a configuration option or even both).
4. always migrate but advise backups: advise users to configure backups in the project/systemd unit documentation and suggest appropriate solutions with os-specific packaging like [`autopostgresqlbackup`](https://packages.debian.org/stretch/admin/autopostgresqlbackup).

knadh: 1 is already the current behaviour. If the binary changes and if it's missing a required DB migration, it exits with a non-zero code and prints a message on stdout prompting the user to run `--upgrade`.

Thinking about it, we can go with the PR (`ExecStartPre`) and advise users to do a DB backup before upgrading the binary by documenting it in the unit file. Eventually, a `--backup` shortcut can be added to make backups easy.

PR #445, opened 2021-08-28
Fix whitespace inconsistency in install scripts

tusharsadhwani:
There was a mix of tabs and spaces in the install shell scripts. This should migrate all of them to tabs.


mr-karan:
Thanks!

PR #463, opened 2021-09-11
UPDATE: Replaced go get(deprecated) with go install

mohit1795:
Replaced `go get` with `go install`.
`go install` builds and stores binary at `gobin` path. So also replaced `stuffbin` path to `gobin`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,8 +8,7 @@ BUILDSTR := ${VERSION} (\#${LAST_COMMIT} $(shell date -u +"%Y-%m-%dT%H:%M:%S%z")
 
 YARN ?= yarn
 GOPATH ?= $(HOME)/go
-GOBIN ?= $(HOME)/gobin
-STUFFBIN ?= $(GOBIN)/stuffbin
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -8,7 +8,8 @@ BUILDSTR := ${VERSION} (\#${LAST_COMMIT} $(shell date -u +"%Y-%m-%dT%H:%M:%S%z")
 
 YARN ?= yarn
 GOPATH ?= $(HOME)/go
-STUFFBIN ?= $(GOPATH)/bin/stuffbin
+GOBIN ?= $(HOME)/gobin
```
knadh, line 11: This is not correct. There is no `$HOME/gobin` directory in the standard installation. By default, it is indeed `$GOPATH/bin`.

mohit1795: No worries if that is not default i will change it to `$HOME/go/bin`

knadh:
Closing this due to inactivity. Already fixed here https://github.com/knadh/listmonk/commit/623030a0c56540eaf9eb10a0bf31d16bcb6e00fc.

PR #576, opened 2021-11-08
Clear placeholder on focus

MickGe:
Because it could be disturbing for people who are not confident with forms not being able to delete words before writing.


MickGe:
*Sorry, I renamed the branch…*

PR #579, opened 2021-11-10
embedded Inter font into binary

sweetppro:
embedded Inter font to avoid calls to Google fonts

only Woff2 fonts have been added, from here (utliising all charsets):
https://google-webfonts-helper.herokuapp.com/fonts/inter?subsets=cyrillic,cyrillic-ext,greek,greek-ext,latin,latin-ext,vietnamese


NicoHood:
wouldnt it make sense to preload those fonts in the html to speed up site loading?

Issue #590, opened 2021-11-16
German (formal) language file


SvenPe:
**Is your feature request related to a problem? Please describe.**
The german language has two pronouns to address people. The informal (Du) and the formal (Sie). The de.json currently implemented uses the informal version which, in some cases, is not acceptable for a newsletter. 

**Describe the solution you'd like**
We made a translation file for the formal version de-formal.json that we think would be useful to be implemented. Furthermore it would probably make sense to adjust the current de.json file to represent the use of the informal pronoun in the following way:
    "_.code": "de-informal",
    "_.name": "Deutsch (de | Du)"

The attached zip contains both json files with the mentioned changes.

[Listmonk_Languagefiles_de.zip](https://github.com/knadh/listmonk/files/7544273/Listmonk_Languagefiles_de.zip)
"

NicoHood:
Makes sense to me to add both! If you create a PR I can review the translations. I recommend first commiting a copy of the current translation and then replacing it with the new one. This diff commit is easier to review then.

knadh:
Thanks for the submission @SvenPe. I don't thin it is ideal to have a formal and informal version of the same language in an application. This is applicable to many languages and following this logic, every multi-lingual app can end up with many different styles for the same language.

I just checked a few apps (including my OS) and formal/informal isn't a pattern that's followed anywhere. We could just edit the existing German language pack to make the language a bit formal, right? The difference between the packs already seem small.

NicoHood:
There is no such thing "a bit formal". It is a common option  to have both variants. But not all software has the effort to support both.

knadh:
> There is no such thing "a bit formal".

Sure there is, depending on the language. There can be a whole spectrum of styles: extremely casual, semi casual, casual, semi formal, very formal, extremely formal ... Multi-lingual apps generally just stick to something close to formal.

> It is a common option to have both variants.

Never came across this pattern. Could you point to some popular multi-lingual apps that have this?

SvenPe:
I do see your point of increased complexity with all the variants @knadh , but see your argument coming a bit short. 
For most apps it will be true, that this is not needed, as the company distributing/offering the app will choose to either go with formal pronoun (like we do), if the target group is more conservative, or with the informal pronoun, if the target group is more modern. Either way the company will adjust this according to their target group. 
[...]

PR #593, opened 2021-11-17
Changed translations to the formal german pronoun "Sie"

SvenPe:
As discussed here https://github.com/knadh/listmonk/issues/590 I have now create a PR to update the translation to be "close to formal"


NicoHood:
I am not sure if formal is a better default. There are pro and cons on both sides. If you ask me, I need informal for my project. So I'd not agree on merging this as a default of course, but I can review the translation itself.

knadh:
hm, the question would be, which style of language here would be fine (does not need to be perfect) for a wide range of German users. Anyone who needs a specific style can always load a custom language pack. We're just trying to have a default that is okay for a large German audience.

Unfortunately, with zero knowledge of the language or its cultural nuances and context, I'm unable to gauge.

SvenPe:
It probably depends on who you @knadh , as a developer, see as your regular user. I wouldn't say that there is either a majority for the formal or the informal pronoun in German businesses. It greatly depends on the industry sector, whether it's a start-up or a mature company etc.. 
From a more objective point of view, the formal approach could be better suited as a default, as you will not "offend" anyone as you would using the informal pronoun, while using the formal pronoun in an informal context would just be a bad fit. On the other hand one could assume that most mature companies (that might be more likely to use the formal pronoun) would use some shareware instead of an open-source solution from Github.

NicoHood:
Exactly. Thatswhy I'd go for both versions, as I already said.

@knadh In german there is only formal and non formal. There is no inbetween, as you might expect. It is just a grammar thing. I think this also applies to french, but I am not an expert there. To me it would make sense to have both, I've seen it multiple times. It depends on if somebody is willing to maintain two languages, but as we've seen there is. And changing a few words is not that hard either.

ChrisTG742:
Do we really need this in the backend?
If so, I'd at least like to have the choice of formal and non-formal and not being forced to use formal german from now on.
So why not creating an additional translation-file instead of changing the already present one?
Just my 2 cents.

knadh:
It looks like we've reached an impasse :)

- I am not in favour of **bundling** different styles of the same language. For German, it may be two styles, but for some other language, it could be N. Once the precedent of bundling different styles of the same language is established, it'll be difficult to review and maintain different styles for different languages. With just two for German, we already don't have consensus.
- There does not seem to be a consensus on the style of German (formal/informal) that should be in the default bundle. We can leave this PR open and maybe wait for input from more German speakers.

In the meanwhile, we can add a section on https://listmonk.app/docs/i18n/ listing as many possible language packs and styles. This section can also be linked directly in the Settings -> General -> Language UI in listmonk so that users can easily discover and use them.

atomtigerzoo:
Hi! In german languages it is just a matter of personal (or business) preference and/or your actual audience in my opinion. Wordpress for example installs the informal (_du_) version as default. You can also install and switch to formal (_Sie_). WooCommerce also uses informal as default and needs to be updated and sometimes edited to match the formal version in all views and mails. This is, for business oriented shops, sometimes tedious to do and check if your audience is more or less business people/professionals.

I would like to have a good choice of both translations in every software that targets the german speaking market, because it has it's use cases - formal and informal needs to be a simple choice IMHO. One setting should be the default, formal or informal - toss a coin :) But the other should be installable or choosable easily for an install.

Recently some buiness newsletters and websites in Germany changed their tonality to informal after years of writing formal. That upsets some readers as they seem to dislike it as it looses professionality in their opinion. Most banks and insurances use formal. Some use informal - that I don't even like. But there might be an audience. As for newsletters and shops, I don't care most of the time. But sometimes I like formal. Sometimes I want informal.

_Long story short:_ it should be an easy switch from one address to the other. And listmonk needs both for the german language.
_Which should be the default?_ Toss a coin and make the other easily to switch to.

HTH 🎈

knadh:
Thanks @atomtigerzoo. listmonk already allows loading any number of custom language packs with the `--i18n` flag, after which they can be easily switched from the UI. The confusion here was about **bundling** multiple styles of the same language into the binary.

I think the best way forward is what I'd suggested earlier. Keep the current informal German language pack bundled in the core. List all additional language packs on the website that can be downloaded and installed. I'll set this up.

samuelT2:
Joke on me - I just translated all missing german terms [here ](https://listmonk.app/i18n/), just to find this pull request... I'll wait until it's merged to see if any terms are still missing.

knadh:
Added a direct link on the UI to [additional language packs](https://listmonk.app/docs/i18n/). @SvenPe, for now, I've linked to de.json in this PR, but it would be great if you could share the latest rebased version. We can host it on listmonk.app via the [listmonk-site](https://github.com/knadh/listmonk-site) repository.

image

bb-Ricardo:
Great example of a bike shed discussion.

I favour informal way more then formal german. Can you create votes on github?

knadh:
I don't think we have a large enough voting pool to be meaningfully decisive.

For now, listmonk bundles one version for all languages (here, informal German as that just happened to be the first community contribution). Custom modifications and styles can be loaded easily with `--i18n-dir`. This is already documented here: https://listmonk.app/docs/i18n/

Also, changing the style of the bundled language will abruptly change it in existing installations after an upgrade which isn't ideal. I left the PR open for someone to confirm the hosting of the formal version somewhere outside of this PR.

bb-Ricardo:
Sorry, then I misunderstood and for hijacking this PR.

BTW: I love listmonk, it is a great tool to maintain our mailing list. You're doing great.

Issue #521, opened 2021-10-18
Email Template Builder


ansidev:
**Is your feature request related to a problem? Please describe.**
Copy and paste HTML code for email template is not easy for non-tech people.

**Describe the solution you'd like**
As a user, I want an email template builder (drag and drop to create it).

ansidev:
I recommend https://github.com/unlayer/vue-email-editor

jackdanielux:
+1 Vue one looks good.

emred2700:
I second MJML. I believe if mjml would be the templating lang for email creation, there will be no need for drag and drop. People then can go either create their emails easily or use Passport(Drag and drop tool on top of MJML from Mailjet) and easily copy and paste the output here.

halms:
I also think that MJML would be a great addition to listmonk, as it makes Email-HTML somewhat bearable.

My ideal workflow would be to have MJML available both in the template and the actual email content editor, possibly both via a graphical editor and as MJML code.

The basic idea is that one could create the template as MJML, and then define one container as the one that holds the content.
The individual component could then just be more MJML that gets inserted in this container and then the whole thing gets rendered to HTML.

In terms of rendering, one would probably have to rely on an external API. MJML is a JavaScript library and there are no Go ports that I know of.
When using a browser, rendering could be offloaded to the client, but that does not apply for using the API to generate Emails.
There one would have to use some MJML-API to render the template. MJML offers a free API for that (https://mjml.io/api), but it also possible to selfhost this with a minimal server around the library (e.g. https://github.com/danihodovic/mjml-server).

I could imagine myself looking into implementing something like this after I familiarize myself with Go a bit.
Except for a change in the data structure, I think the implementation (at least for MJML code) could be rather straightforward.

A visual editor for MJML (most probably the above-mentioned grapesjs-mjml) could then be put on top. It is just not yet clear to me how this would work for the template-and-content workflow.

One comment on the recommended: unlayer/vue-email-editor
It seems to me that unlayer is a commercial product, albeit with a free version available. The Vue component is just a wrapper that includes this and loads the editor from unlayer.com
I don't know what limitations are there if only embedding anonymously without signing up, but it definitely is an externally loaded, non open-source dependency.

aksanoble:
+1 to the need for the email template builder. 
[...]

PR #596, opened 2021-11-19
Add MJML Support

natekfl:
This adds support for [MJML](https://mjml.io/). Somewhat addresses #521

As this does require the MJML CLI to be installed globally on the system (or docker container), this does somewhat break the "Single binary app" idea of listmonk. However, if the MJML CLI is not installed globally, it will simply log it in the console, and all MJML options will be hidden from the admin ui.

No worries if you feel this change is out of scope for this project due to the above.

Changes:
- Templates now have a type, which can either be HTML or MJML
- Campaign bodies have the new content type MJML. This can only be selected if the template for the campaign is also MJML
- MJML templates compile to HTML, including comment breakpoints for Outlook support
-  Added `static/email-templates/default.mjml.tpl`. While not currently used in the source code, this replicates ` static/email-templates/default.tpl` but using MJML, for reference.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,6 @@
 FROM alpine:latest
-RUN apk --no-cache add ca-certificates
+RUN apk --no-cache add ca-certificates nodejs npm
+RUN npm install -g mjml
 WORKDIR /listmonk
 COPY listmonk .
 COPY config.toml.sample config.toml
--- a/cmd/admin.go
+++ b/cmd/admin.go
@@ -12,12 +12,13 @@ import (
 )
 
 type serverConfig struct {
-	Messengers   []string   `json:"messengers"`
-	Langs        []i18nLang `json:"langs"`
-	Lang         string     `json:"lang"`
-	Update       *AppUpdate `json:"update"`
-	NeedsRestart bool       `json:"needs_restart"`
-	Version      string     `json:"version"`
+	Messengers    []string   `json:"messengers"`
+	Langs         []i18nLang `json:"langs"`
+	Lang          string     `json:"lang"`
+	Update        *AppUpdate `json:"update"`
+	NeedsRestart  bool       `json:"needs_restart"`
+	Version       string     `json:"version"`
+	MJMLSupported bool       `json:"mjml_supported"`
 }
 
 // handleGetServerConfig returns general server config.
@@ -53,6 +54,8 @@ func handleGetServerConfig(c echo.Context) error {
 	out.Update = app.update
 	app.Unlock()
 	out.Version = versionString
+
+	out.MJMLSupported = mjmlSupported
 
 	return c.JSON(http.StatusOK, okResp{out})
 }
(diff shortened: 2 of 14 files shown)


knadh:
Thanks for this, but you're right, `exec()`ing an external program and adding it as dependency to the core isn't ideal. I don't think we should merge this.

That said, MJML compilation should only be done once when loading the campaign / template. Here, if you have a million subscribers, mjml will be executed a million times. You should move this to maybe `manager.addCampaign()`

 https://github.com/knadh/listmonk/blob/db0bdb0a4e0fcbcd3f95074ab9eb64e250a7feec/internal/manager/manager.go#L679

natekfl:
> You should move this to maybe `manager.addCampaign()`

Thanks for the tip there. It gets a bit weird with the order of compilation, as if the user uses go template conditionals, we want that to run on the MJML, and then the final MJML gets compiled to HTML. Compiling the MJML to HTML and then doing the go templating would result in undefined behavior for anything except just inserting variable values.

Anyway, as was discussed in #521, since there are no Go ports of MJML (at least that I know of), the only way to implement it outside of calling the global CLI tool is to use an MJML api. For that you could either use the Mailjet public api (https://mjml.io/api), which carries with it it's own privacy/security concerns, or run one on the machine (e.g. https://github.com/danihodovic/mjml-server), at which point you're basically doing the same thing as installing MJML globally. So the TL;DR is that there really isn't a better way of doing it that I know of, and like I said if that's out of scope for this project no worries, thanks anyway!

knadh:
> Compiling the MJML to HTML and then doing the go templating would result in undefined behavior for anything except just inserting variable values.

Ah, I assumed this would be the way to do it, where MJML sits on a layer above native HTML/Go templating. MJML has a problem with `{{ the go templating syntax }}`, is it?

> is that there really isn't a better way of doing it that I know of

Looks like until a Go port of mjml is available, a seamless integration wouldn't be possible. Users will have to write mjml elsewhere and copy-paste it as HTML.

natekfl:
> MJML has a problem with `{{ the go templating syntax }}`, is it?

It's not that it directly has a problem with the go template syntax.

For example, if you have 

```xml
<mjml>
    <mj-body>
        <mj-section>
{{if condition}}
             <mj-column>
                 Column 1
             </mj-column>
{{end}}
             <mj-column>
                 Column 2
             </mj-column>
        </mj-section>
    </mj-body>
</mjml>
```

If you run go templating first then MJML, it will work as expected, where column 1 is only present if the condition is true.
However, if you run MJML first, then it will compile to html as if both columns are present, but one of the column `td`s will be surrounded by the go template conditional. However the second column will have the styling and breakpoints as though there are two columns, regardless of if the first column ends up being present.

Anyway, like you said this is all pretty moot since there are no good ways to include mjml in listmonk at the moment. We are currently using this fork in production, even with the subpar method, because MJML is a requirement for us. However if somehow in the future a golang port of MJML is created, we can revisit this :)

knadh:
> Anyway, like you said this is all pretty moot since there are no good ways to include mjml in listmonk at the moment. We are currently using this fork in production, even with the subpar method, because MJML is a requirement for us. However if somehow in the future a golang port of MJML is created, we can revisit this :)

Yep. Thanks for the PR again. Closing this for now.

PR #605, opened 2021-11-29
Add Dutch (nl) translation

Jjagg:
This adds a Dutch translation.

Similar to German, in Dutch there are formal and informal pronouns (#593).
I used informal pronouns, because that's what I need for my use-case. Culturally, I think for Dutch folks this is not as contentious and by far most users will want to use informal pronouns.


knadh:
Thanks @Jjagg!

Issue #650, opened 2022-01-05
View and click details by subscribers


mrhydejc:
**Is your feature request related to a problem? Please describe.**
I need to know which subscribers open/click a campaign. (linked to https://github.com/knadh/listmonk/issues/473)

**Describe the solution you'd like**
A CSV export could do the job.
When Individual subscriber tracking is enable, add an export button in analytic view, just below the view's graph and the click's graph.
![export-details](https://user-images.githubusercontent.com/11319259/148187144-c9a8f7d3-e502-4e08-b861-3f3a91c90e18.png)
View export :
![view-export](https://user-images.githubusercontent.com/11319259/148187203-1ead49b5-ba5f-4729-b2e6-c2dd4ad208c7.png)
Click export :
![view-click](https://user-images.githubusercontent.com/11319259/148187241-9f0fcde8-e2b7-4b36-ba35-f85331cef50f.png)

charlesmudy:
+1 for this, the overall views and clicks is great but no context to it, knowing the individual subscribers that clicked, viewed a campaign will be great

charlesmudy:
@knadh @mrhydejc this is what I managed to do and looks quite ok. 

`
WITH camps AS (
SELECT
			subscriber_id,
			COUNT(CASE WHEN subscriber_id = null THEN 0 ELSE 1 END) AS views 
FROM campaign_views
WHERE campaign_id = 586
GROUP BY subscriber_id ),

links AS (
SELECT
			subscriber_id,
			COUNT(CASE WHEN subscriber_id = null THEN 0 ELSE 1 END) AS clicks 
FROM link_clicks
WHERE campaign_id = 586
GROUP BY subscriber_id )

SELECT subscribers.id, subscribers.email, camps.views, links.clicks FROM camps, links, subscribers
WHERE camps.subscriber_id = links.subscriber_id
AND camps.subscriber_id = subscribers.id
AND links.subscriber_id = subscribers.id
`


<img width="761" alt="Screenshot 2022-04-07 at 13 14 25" src="https://user-images.githubusercontent.com/1445007/162196482-b153f47f-1c06-460a-b7f0-994fc8f89a4c.png">

mrhydejc:
@charlesmudy take a look at this PR : https://github.com/knadh/listmonk/pull/689

niraj-ranjan-comprinno:
Hi, Above query is failing in scenario like
Suppose no one clicked the link but view the compaign

c-seeger:
[...]

PR #660, opened 2022-01-12
in analytics, add an export button above the links graph

mrhydejc:
On analytics page, add an export button above the links graph to export links's clicks details. #650
This information is not otherwise available.
Tested on my cluster. 
This is my first experience in golang and vuejs. I hope I have followed the rules.


mrhydejc:
it's look like this.
image

knadh:
Thanks for the PR @mrhydejc 

>  This is my first experience in golang and vuejs. I hope I have followed the rules.

I took a quick look and it looks like a surprisingly good PR for a first time Go / Vue commit! Will review properly over the weekend and submit comments.

mrhydejc:
Thanks @knadh 
I plan to make export for the others analtytics categories (view, click, bounce).
Would you like I push them with this PR or with separates PR.

About i18n, I duplicated the key `subscribers.export` to `campaigns.analytics.export`.
Shouldn’t I be merging them in a new key `globals.buttons.export` ? (Be careful of language packs)

knadh:
Since it's one feature on the same page, it can be in the same PR. Yep, it should ideally be moved to `globals.buttons.export`.

knadh:
- `handleGetCampaignAnalyticsExport()` duplicates much of what `handleGetCampaignViewAnalytics()` already does. Querying DB for different types of stats, scanning results to a struct, sanitizing ... A better way would be to move this logic to a shared `getCampaignAnalyticsData()` or something and reuse that in the current JSON handler and in the new CSV export handler. Instead of adding new SQL queries, the existing queries can be reused (with some modifications to include any new fields required by the CSV export).

- `campCountStats` can be renamed to something a bit more generic like `campStatsData`, and the additional fields required by the CSV export can also be added there. Even if these new fields show up in the JSON API, it should be fine, but these fields can be selectively turned-off for JSON export by using `json:"omitempty"`. This will eliminate a lot of type selection logic and reflection.

- The bounce export seems to be exporting subscriber data and not just aggregates. I think that's more appropriate on the Subscribers -> Bounces page.

mrhydejc:
Hi @knadh, thanks you for the review.
Using  single model with `json:"omitempty"` is a better approach !

Actually, my initial need is to export subscribers views and clicks with their timestamp.
As "Susbribers > Views" and "Suscribers > Clicks" pages does not exists, I thought it would be nice if the statistics CSV exports were not aggregated. It meets my need. I did the other ones (click + bounce) to make the feature consistent. (one export for each graph).
ATM, each export is not aggregated except click. (otherwise it would be the same as link's export)

With exports not aggregated, it could be difficult to share SQL as :
- There is `GROUP BY` for graphs and not for the CSV
-  CSV's queries use `OFFSET/LIMIT` as there is lot more records.
It’s probably could be done, but should we?

Do you want exports on statistics page to be aggregated (same datas as the graph) ?
If yes, should I : 
- Update export on analytics pages with aggregated queries
- Create "Suscribers > Views" page and put the not-aggregated export on it (new PR)
- Create "Suscribers > Clinks" page and put the not-aggregated export on it (new PR)
- Move the not-aggregated Bounce export on "Suscribers > Bounces" (new PR)

knadh:
Apologies for the super delayed response @mrhydejc. Wasn't able to find time to review the big PR in detail.

Just took a close look at this and also ran it locally on a large DB. It's almost there! Couple things:

- `2022/01/31 21:52:39 campaigns.go:735: error executing query: sql: Scan error on column index 2, name "suscriber_email": converting NULL to string is unsupported` (when exporting clicks data)
- Could you resend the PR from a feature branch from your repo instead of `master`? That way, any final (minor) cosmetic tweaks I'd like to add to the PR can be merged alongside your commits.

mrhydejc:
continue on https://github.com/knadh/listmonk/pull/689

PR #668, opened 2022-01-19
Spiffy up containerized dev stack

avanier:
This PR suggests a few minor changes that should make the containerized dev stack a bit easier to start.

In it's current state the stack will error out on the `config.toml` file being absent, and that message easily gets buried under the rather verbose logs produced by the frontend stack. Likewise, initializing the database did not reliably work.


mr-karan:
@avanier Thanks for this.

I always had `dev/config.toml` in my local clone but looks like the global `.gitignore` had omitted it, hence the error you faced.

Apart from that, changes LGTM :+1: 

cc @knadh

PR #680, opened 2022-01-26
Display unique campaign views instead of overall views

### LINKED ISSUE #676: Display unique campaign views instead of overall views
image

image

Currently, the UI displays the total sum of views per campaign. If we have 10 users and only one user opens the email 10x times, the views' percentage will show 100%. 

This is misleading (#522, #561, #571) and in my opinion rather useless. The number displayed should be 1 view and 10% open rate. 

I could even try to create a PR for this, but first need your approval @knadh. I've seen there is a work in progress on adding another field #572, but imo this should be replaced.

marcinkunert:
I haven't tested it on a huge database, but it should do the trick. As far as I know, no other changes should be required.

Closes #676

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,3 @@
 frontend/* linguist-vendored
 VERSION export-subst
+* text=auto eol=lf
--- a/queries.sql
+++ b/queries.sql
@@ -571,7 +571,7 @@ WITH intval AS (
     -- For intervals < a week, aggregate counts hourly, otherwise daily.
     SELECT CASE WHEN (EXTRACT (EPOCH FROM ($3::TIMESTAMP - $2::TIMESTAMP)) / 86400) >= 7 THEN 'day' ELSE 'hour' END
 )
-SELECT campaign_id, COUNT(*) AS "count", DATE_TRUNC((SELECT * FROM intval), created_at) AS "timestamp"
+SELECT campaign_id, COUNT(DISTINCT subscriber_id) AS "count", DATE_TRUNC((SELECT * FROM intval), created_at) AS "timestamp"
     FROM campaign_views
     WHERE campaign_id=ANY($1) AND created_at >= $2 AND created_at <= $3
     GROUP BY campaign_id, "timestamp" ORDER BY "timestamp" ASC;
--- a/queries.sql
+++ b/queries.sql
@@ -581,7 +581,7 @@ WITH intval AS (
     -- For intervals < a week, aggregate counts hourly, otherwise daily.
     SELECT CASE WHEN (EXTRACT (EPOCH FROM ($3::TIMESTAMP - $2::TIMESTAMP)) / 86400) >= 7 THEN 'day' ELSE 'hour' END
 )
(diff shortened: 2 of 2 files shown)


.gitattributes
```diff
@@ -1,2 +1,3 @@
 frontend/* linguist-vendored
 VERSION export-subst
+* text=auto eol=lf
```
NicoHood, line 3: This should not go into this PR I guess?

marcinkunert: yeah, without it I was unable to clone the repo in widows. If you want i could create a new PR, The change is minimal so it should be doable using the github build-in editor

knadh:
I'll test this locally on a large DB soon and then merge.

NicoHood:
I think we should merge this PR/add this feature.

You mentioned a good point about individual subscriber tracking. I am personally on a privacy friendly side. However what is the point, if the graphics are totally useless (they are) if no individual people are tracked. Maybe it would make sense to USE or NOT TO USE the tracking at all. Is that maybe a more user friendly experience idea?

knadh:
> However what is the point, if the graphics are totally useless (they are) if no individual people are tracked.

I disagree. With individual tracking off, the PR renders a flat graph. There are thousands of actual views recorded (without individual subscriber information), including duplicate views, granted, but it is still valid information. To show 0 in such cases is semantically incorrect.

image

I think the better approach is to show unique views (like this PR does) when individual views are turned on, and non-unique views when they're off. This non-unique graph, despite including an unknown number of duplicates, still conveys some useful information on trends.

image

I'll hold off on this PR and push a patch that does unique/non-unique conditionally based on the privacy setting. It'll need the existing SQL queries to have string interpolation based on the privacy setting (there's no way to do a conditional DISTINCT directly) so requires a bit more work.

NicoHood:
But then the UI should also reflect this and should not show any percent, like it does at the moment.

knadh:
Yep, when subscriber tracking is off and the graph's non-unique, the % dials will not render.

marcinkunert:
@knadh 

> I'll hold off on this PR and push a patch that does unique/non-unique conditionally based on the privacy setting. It'll need the existing SQL queries to have string interpolation based on the privacy setting (there's no way to do a conditional DISTINCT directly) so requires a bit more work.

Maybe instead of creating a string interpolation query, just fetch both of these values. I'm not sure though how this affects the performance.

```SELECT campaign_id, COUNT(DISTINCT subscriber_id) AS "count", COUNT(*) AS "total_count", DATE_TRUNC((SELECT * FROM intval), created_at) AS "timestamp"```

After that (depending on tracking settings) one of these counts could be returned to REST endpoint.

knadh:
@marcinkunert 2614b072f2a20f7d9d32a30bd5da54c902948b18 addresses this. The queries are string interpolated based on the privacy setting just once and are prepared and kept on app boot. Having 4 big queries and executing them in app logic would've been complicated.

marcinkunert:
@knadh sure, looks good!

Issue #627, opened 2021-12-16
Expose metrics for external consumption


avanier:
**As a** listmonk administrator,
**I want to have** metrics available externally
**So that I can** consume them and implement things like alerting with outboard systems.

<hr>

### Context

In our particular deployment of listmonk we're processing bounces from SES and the relevant subscribers are not being blocklisted automatically. Exporting metrics that we could hook into our monitoring system would help bring issues to our attention as they arise.

### Misc

I'm specifically thinking of opening a PR to expose Prometheus metrics. I've seen that some of the stuff is already instrumented with `go-metrics` for internal dashboards, so building on that should be no biggie.

Do note, I'm suggesting exposing Prometheus metrics since I'm used to the tooling but we could just as well do something else like Influx, or Statsd. Whichever the community feels is the best common denominator.

knadh:
Hi @avanier, what kind of metrics are you proposing? Can you give some examples?

Prometheus format should be fine.

avanier:
For my particular use-case I know could use exporting something like `listmonk_bounces_processed` or something along those lines. Otherwise I'm guessing messenger / mailer errors? So generally I'm thinking of things representative of _"things that you can do something about and that waiting to discover wouldn't be great"_.

The Go Prometheus also has a bunch of _"free"_ ones like `build_info` that's pretty handy in Kubernetes to watch your deployments roll out.

I'm open to suggestions if there anything that you feel should be part of a first pass.

knadh:
Makes sense. Off the top of my my head:

- Bounce counts
- Campaign send rates for running campaigns
- Campaign counts grouped by status?
- Subscriber counts grouped by status? (can be an expensive query).

Many of these queries already exist [here](https://github.com/knadh/listmonk/blob/13edf426a6f840610d0b32f4f128dc8f67364240/queries.sql#L791) and [here](https://github.com/knadh/listmonk/blob/13edf426a6f840610d0b32f4f128dc8f67364240/queries.sql#L823).

avanier:
Hey, status update with good new and bad news.

Good news: I got started and it's looking good. You can check the branch I've got going on [here](https://github.com/knadh/listmonk/compare/master...avanier:feature/add-prometheus-metrics).

[...]

PR #694, opened 2022-02-09
Add Prometheus metrics

avanier:
This pull request implements #627 .

The goal here is to expose some basic metrics over a Prometheus endpoint.

~This is very much a draft PR for now. I would expect this to get rebased into something cleaner before it's taken out of draft state.~ Undrafted, yay! \o/


cmd/admin.go
```diff
@@ -8,7 +8,7 @@ import (
 	"time"
 
 	"github.com/jmoiron/sqlx/types"
-	"github.com/labstack/echo/v4"
+	"github.com/labstack/echo"
```
mr-karan, line 11: I don't think this should be changed.

> Any of these versions will allow you to import Echo as github.com/labstack/echo/v4 which is the recommended way of using Echo going forward.

https://github.com/labstack/echo

cmd/handlers.go
```diff
@@ -170,6 +172,14 @@ func initHTTPHandlers(e *echo.Echo, app *App) {
 
 	// Public health API endpoint.
 	e.GET("/health", handleHealthCheck)
+
+	if app.metrics.Prometheus.Enabled {
```
mr-karan, line 176: We can just call it `app.metrics.Enabled`

cmd/handlers.go
```diff
@@ -170,6 +172,14 @@ func initHTTPHandlers(e *echo.Echo, app *App) {
...
+		// Public metrics endpoint
+		g.GET(app.metrics.Prometheus.Path, metrics.HandlePromMetrics())
```
mr-karan, line 178: `app.metrics.Path`

cmd/main.go
```diff
@@ -102,6 +107,14 @@ func init() {
 		os.Exit(0)
 	}
 
+	// Set some configuration defaults
```
mr-karan, line 110: This should be a self contained function. `initMetrics` inside `init.go`

cmd/main.go
```diff
@@ -102,6 +107,14 @@ func init() {
...
+	ko.Load(structs.Provider(
+		metrics.PrometheusConfig{
```
mr-karan, line 112: I think the correct way here is to use `ko.Unmarshal`, like done here: https://github.com/knadh/listmonk/blob/master/cmd/init.go#L239-L241

cmd/main.go
```diff
@@ -163,6 +176,12 @@ func main() {
 		messengers: make(map[string]messenger.Messenger),
 		log:        lo,
 		bufLog:     bufLog,
+		metrics: metrics.MetricsConfig{
+			Prometheus: metrics.PrometheusConfig{
```
mr-karan, line 180: You can remove the nested `Prometheus` struct here. Make `Enabled` and `Path` as top level fields to `MetricsConfig`.

config.toml.sample
```diff
@@ -23,3 +23,8 @@ ssl_mode = "disable"
 max_open = 25
 max_idle = 25
 max_lifetime = "300s"
+
+# Prometheus Metrics.
+[metrics.prometheus]
```
mr-karan, line 28: `[metrics]` should be enough

config.toml.sample
```diff
@@ -23,3 +23,8 @@ ssl_mode = "disable"
...
+enabled = true
+path = "/metrics"
```
mr-karan, line 30: Perhaps a default value should be `/api/metrics` to conform with existing URI pattern

internal/bounce/bounce.go
```diff
@@ -145,6 +146,15 @@ func (m *Manager) runMailboxScanner() {
 
 // Record records a new bounce event given the subscriber's email or UUID.
 func (m *Manager) Record(b models.Bounce) error {
+	metrics.Counter{
```
mr-karan, line 149: Instead of defining a Counter object here, IMHO a better and simpler API would be:

```go
metrics.Increment("label")
```

You can then define a `.Increment` function in package which does:

```

// Increment the counter for the corresponding key.
// This is used for Counter metric type.
func (s *Manager) Increment(label string) {
	s.metrics.GetOrCreateCounter(label).Inc()
}
```

What do you think?

internal/metrics/handler.go
```diff
@@ -0,0 +1,77 @@
...
+
+import (
+	"strings"
+
+	"github.com/VictoriaMetrics/metrics"
+	"github.com/labstack/echo"
+)
+
+type MetricsConfig struct {
+	Prometheus PrometheusConfig
```
mr-karan, line 11: `PrometheusConfig` can go away

internal/metrics/handler.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+type PrometheusConfig struct {
+	Enabled bool   `koanf:"enabled"`
```
mr-karan, line 15: These fields can come directly under `MetricsConfig`

internal/metrics/handler.go
```diff
@@ -0,0 +1,77 @@
...
+	Path    string `koanf:"path"`
+}
+
+func HandlePromMetrics() echo.HandlerFunc {
+	return echo.HandlerFunc(
+		func(c echo.Context) error {
+			metrics.WritePrometheus(c.Response().Writer, true)
```
mr-karan, line 22: Whether to expose Go process metrics or not, can also come as config instead of passing a hardcoded `true`.

mr-karan (changes requested):
I've gone through the PR and yeah this would require a bit of design changes. Overall the approach looks fine. Needs some more thought in designing the API for the internal metrics package, I suggested a simpler approach please have a look.

I've not yet gone through the Echo middleware part, shall do so once these changes are finalised.

internal/metrics/handler.go
avanier: Oh, you mean we always register the handler, but it it's disabled it's just an empty handler?

That would be nicer to read.

config.toml.sample
avanier: I'm curious of your thoughts. Prometheus [best practice](https://prometheus.io/docs/introduction/first_steps/#configuring-prometheus) expects this path to be at `/metrics` and expects it to be available unauthenticated.

This is convenient and makes a lot of sense in environments like Kubernetes where reverse-proxying and isolating resources is trivial _(within the realm of you running Kubernetes)_.

I don't get a sense of what the average Listmonk deployment is like, so that might not be a good idea depending on our target audience. Especially the part where metrics are available unauthenticated.

internal/bounce/bounce.go
avanier: I'm not sure I understand the suggestion.

Would the idea be to pre-register all the different metric names in the Manager and give them the appropriate methods through an interface and just pass the appropriate label map when they're invoked?

config.toml.sample
knadh: Path should not be configurable at all. Best to stick to one universal metrics path in all listmonk installations to avoid future confusion (and support burden).

internal/bounce/bounce.go
avanier: I'm re-reading and checking how this could be implemented. I don't think it's possible since every combination of metric name and label set constitutes a unique counter, and that can't be pre-declared before I know what labels will be provided

cmd/admin.go
avanier: Fixed, my package woes was caused by some bad copy-pasta. :spaghetti:

internal/bounce/bounce.go
avanier: I just gave it a try. I think it could be written, but I'm fairly certain we'll be clobbering our object and create races.

I think it could look something like this.

```go
package "metrics"
// [...]

func (c Counter) New(n string) Counter {
	return Counter{
		Name: n,
	}
}

func (c Counter) IncrementLabels(l Labels) {
	c.Labels = l
	c.Inc()
}

func (c Counter) Increment() {
	metricName := composeMetricName(c.Name, c.Labels)
	metrics.GetOrCreateCounter(metricName).Inc()
}
```

But my spider-senses are more than tingling when I look at this. :grimacing:

mr-karan: Ah, sorry for not being clear in the first time. I don't think we need to register labels before hand. Here's a complete example of what I had in mind:

In the metrics internal package:

```go
// Manager represents options for storing metrics.
type Manager struct {
	metrics         *metrics.Set
	namespace       string // Optional string to prepend all label names.
}

// NewStats returns a new configured instance of Manager.
func New(ns string) *Manager {
	return &Manager{
		metrics:   metrics.NewSet(),
		namespace: ns,
	}
}

// Increment the counter for the corresponding key.
// This is used for Counter metric type.
func (s *Manager) Increment(label string) {
	s.metrics.GetOrCreateCounter(s.getFormattedLabel(label)).Inc()
}

// Decrement the counter for the corresponding key.
// This is used for Counter metric type.
func (s *Manager) Decrement(label string) {
	s.metrics.GetOrCreateCounter(s.getFormattedLabel(label)).Dec()
}

// Set updates the key with a float64 value.
// This is used for Gauge metric type.
func (s *Manager) Set(label string, val float64) {
	s.metrics.GetOrCreateFloatCounter(s.getFormattedLabel(label)).Set(val)
}

// getFormattedLabel prefixes the label with namespace (if non empty).
func (s *Manager) getFormattedLabel(l string) string {
	if s.namespace != "" {
		return s.namespace + "_" + l
	}
	return l
}
```

So, we define all the ops related to metrics exposition inside the package. This way, the API to use this package becomes cleaner.

NOTE: Since we are using `GetOrCreate` functions, we don't need to register any metrics before hand either. This makes it simpler to add ad-hoc metric names.

In the place where you need to capture a metric (i.e. bounces etc):

```go
var (
	labelBouncesProcessed         = `bounces_processed`
)
app.metrics.Increment(labelBouncesProcessed)
```

This will emit `listmonk_bounces_processed` (`listmonk` being the `namespace`) which is a valid Prometheus metric. In case you need to add labels to the metric, you can use a simple string concat `bounces_processed + "{" + label_key + "=" + label_val + "}"` (`fmt.Sprintf` can be used for readability and I don't think that should be a problem here performance wise).

Hope that makes sense. Let me know if this approach sounds easier to implement.

mr-karan: Here's an example of how to add dynamic labels using string concat: https://github.com/zerodha/fastglue-metrics/blob/main/metrics.go#L98

avanier: Yup, that makes sense to me wno. Also, that fastglue middleware gives a good exmaple of how to get request durations! I'm still early in digging into Echo middlewares. Hopefully I might be able to get that in too.

Yeah, with those examples I should be able to put it together. :pray:

avanier:
Almighty, I think this might be able to come out of draft now.

I believe this should be in a better form than previously and closer to what you had in mind.

One thing you'll note is that I did not go with the string concat to declare the metric name and labels inline. Hopefully, this is an acceptable compromise. I understand how the string concat gets it out of the way, line-count wise, but while metric names _are_ technically labels at storage time on most backends, they're never referred as such in the Prometheus or VictoriaMetrics codebase. I felt that keeping the reference to the metric name being separate from the labels was worthwhile for _"standards adherence"_.

Either way, if this doesn't pass muster, it's not a big refactor. I'm hoping this is unobtrusive enough to slip by. :smile:

_(Also, this is a mess of commits. I might rebase that into something better organized.)_

avanier:
@knadh && @mr-karan . May I bump for review?

mr-karan:
@avanier Firstly, excellent work so far :) Most of the things are in-place, I just had to do some minor refactoring on how the config is loaded, function/variable naming and removing some redundant code that I found.

I was not really sure if I can push the changes to your fork (if you give write access I can maybe). For now, I've pushed changes in this repo itself and you can see what is changed: https://github.com/knadh/listmonk/commit/bc7acfe21301b459f08c2bf5c8f20f220143a43d. Please let me know if the proposed changes are fine.

I've some minor concerns before we land this in upstream:

1) Do we keep `metrics` config in `config.toml` (or in the UI)? If inside `config.toml` is fine, then when users upgrade, they won't necessarily have this config block. To accomodate for that, I've loaded a default config inside app: https://github.com/knadh/listmonk/blob/feature/add-prometheus-metrics/cmd/init.go#L317-L320. This should be fine IMO.

2) Over here, https://github.com/knadh/listmonk/blob/feature/add-prometheus-metrics/internal/metrics/http.go#L60-L63 we are using `reflection` to see if the handler's function type is indeed a Not Found handler. Since the type is function, I guess there's no other way to check apart from using `reflect`. But this check will happen to each and every request, having a `reflect` op on every single incoming request, **may not be a good idea**.  @knadh What do you think?

 Apart from this, mostly looks LGTM! @knadh can do a final review once we discuss the above points. 
 Thanks!

knadh:
1) It should be in the UI perhaps under `Performance`? `config.toml` only has config that's required for the app to boot.

2) Yeah, it's not ideal to reflect on every single request. Can't we just do `response.status == 404` or something? Also, why should 404 be excluded from metrics?

mr-karan:
> It should be in the UI, probably under Performance. config.toml only has config that's required for the app to boot.

Okay, @avanier we need to take that in consideration in this PR. I can help out with frontend changes if you want.

> Also, why should 404 be excluded from metrics?

We're not excluding 404s from the metric. Since we are embedding the HTTP handler path inside metrics, there's a chance of increasing high cardinality (which Prometheus or any time series DB doesn't like - as it makes things slower for them).

For eg, on visiting `/api/abc`, the metric will be like: 

`requests_total{path="/abc", status="404"}`

There could be millions of requests to random paths that don't exist, but they will end up creating unique metric labels (each label is a basically a unique time series). This would have downstream effects on Prometheus/metrics collection server. So, what the intended behavior should be, is we define a hard-coded string (like `/not-found`) and use it for any kind of path value which is 404.

`requests_total{path="/not-found", status="404"}`

> Can't we just do response.status == 404 or something?

Hm, yeah probably we can. This should work I think.

avanier:
@mr-karan I gave this a look. Many thanks for the edits! :pray: I've cherry-picked https://github.com/knadh/listmonk/commit/bc7acfe21301b459f08c2bf5c8f20f220143a43d into this branch. Normally you should have been able to push as I did enable the `Allow edits by maintainers` :heavy_check_mark: on the PR. ¯\\\_(ツ)\_/¯  

That said, I do have a few questions:

* From what I can see in your edits, this would make Prometheus metrics enabled by default. Most applications I've seen always implemented a way to disable metrics altogether. I don't see any issues with this being enabled by default.
* I see you're removed all the code relating to `_info` metrics. Those are pretty useful to have in Kubernetes environments to have a trace of deployments rolling out or correlate version changes to other issues. Would you mind if I brought that back in? Did you see any issues I could address with how that code was written?
* I see you've made the metrics namespace of `listmonk` a configurable option instead of a hardcoded value. Is there a specific reason for this? Normally if a user had multiple Listmonks running, standard Prometheus doctrine would have you scrape it while adding additional labels _(in Kubernetes this would be the default behaviour)_. Having different metrics namespaces would make writing queries aggregating multiple Listmonks hairy in the best of scenarios. It also brings documentation challenges since this would have to be factored in whenever we refer to those metrics. I don't think we'd be doing users a service by making this configurable.

As for the reflection thing I've amended that to a simple `404` check. I did a quick double check and as far as I can tell the end result should be identical.

@knadh Regarding the presence of these settings in the UI, I would personally keep it only in the config file. Considering there are no Prometheus solutions out there that can be configured by clicking, I don't think we'd be leaving too many users of that ecosystem stranded. Either way, I would argue this could be implemented separately if needed?  _(That and my Vue.js is more than rusty ... :sweat_smile: )_

knadh:
> @knadh Regarding the presence of these settings in the UI, I would personally keep it only in the config file. Considering there are no Prometheus solutions out there that can be configured by clicking, I don't think we'd be leaving too many users of that ecosystem stranded. 

While this is subjective, as I said, listmonk only has config it requires to boot in the TOML file. That's a simple standard that can be followed or we end up in situations like this where every setting requires deliberation on where it should go.

@mr-karan @avanier if the API is behind the BasicAuth API key, then it can just be on all the time maybe.

mr-karan:
> From what I can see in your edits, this would make Prometheus metrics enabled by default. Most applications I've seen always implemented a way to disable metrics altogether. I don't see any issues with this being enabled by default.

That's correct. I've implemented in such a way that if the config block isn't present (which it won't be for people upgrading), reasonable defaults are loaded and Prometheus metrics are exposed.

I don't think having an option to disable metrics would make a significant difference, since anyway we are collecting metrics inside the function blocks. Disabling them would simply disable exposing the API Handler/Response. Don't think it's a big deal, anyway it's an internal API.

> if the API is behind the BasicAuth API key, then it can just be on all the time maybe.

Yep, it'll be on all the time as discussed ^ .

> I see you're removed all the code relating to _info metrics. Those are pretty useful to have in Kubernetes environments to have a trace of deployments rolling out or correlate version changes to other issues

Hm, seconded. Although I did had a chat about this with @knadh about exposing metrics like build/version info. Exposing that metric on every single request seemed wasteful. We can reconsider it though.

> I don't think we'd be doing users a service by making this configurable.

Fair point, I agree with you. We can remove it from config, make it hardcoded as a constant in the app as `listmonk`.

avanier:
Alrighty, I'll get on making the changes.

I started having a look and the Vue part is more straightforward than I expected. That said, I'm still facing significant challenges understanding how `koanf` is used.

Right now, my understanding of how the settings API and `koanf` should be interacting. _(Excuse the crude diagram)_

```mermaid
sequenceDiagram
  participant main as main.go
  participant init as init.go
  participant koanf as koanf
  participant config as config file
  participant postgres
  participant environment
  participant request as HTTP request

main->>init: init() 
init->>config: initConfigFiles()
config->>koanf: set config values from config file
init->>environment: [...] initConfigFiles()
environment->>koanf: set config values from environment
init->>postgres: initDB()
postgres->>koanf: set config values from database

main->>init: main()
init->>koanf: initConstants() queries koanf
koanf->>init: returns config values<br />this is where the app settings will be maintained for the API
main->>init: initMetrics()
init->>init: metrics module initialized

main->>main: Application is happy and runs

request->>main: request to /api/settings<br />this actually happends in handlers.go ... but this graph is big enough
main->>request: return App constants as JSON
```

Right now I've pushed c5b92efbe216d8a112c6d31231b56695be6bf6e1 to a separate branch. If I query `/api/settings` everything I've set is returning `false`, most likely because `nil` is being evaluated falsey. I'm somehow expecting `koanf` to do _"something"_ with the app constants so that the right values are returned at query-time &hellip; but that assumption is obviously incorrect on my part. @mr-karan I could use pointers on how to get this to work correctly if I may ask. :pray: 

> Sidenote: I love how the app `SIGHUP`s itself when settings are updated, that's a really clean and simple way to handle this I hadn't seen before. :metal: 

Once I have the `/api/settings` part working correctly, _"I think"_ I should be fine with putting the front end part together.

knadh:
Ah, I thought we weren't doing config or UI settings but planning to keep metrics always on behind the standard BasicAuth API.

avanier:
Errrr ... You had me under the impression that UI settings for any exposed setting.

Right now Prometheus would export the following:

* Select bespoke metrics like bounces, ship-rate, or other things we manually implement as always-on
* HTTP metrics for all routes through an Echo middleware as a toggle-able
* Golang process metrics as a toggle-able

If we make none of those things configurable ... that would solve the all those issues >_< .

I wouldn't think of the Golang metrics as too expensive to collect and export, but few of them are actually very useful in production unless you're a Listmonk contributor who knows how to do Golang performance optimization.

At-large HTTP metrics are usually the bread and butter of Prometheus queries, but they run the risk of being high cardinality depending on the size of your deployment. _(Also high cardinality depends on which Prometheus backend you use, so YMMV.)_ Either way, they're generally high-ish value.

The bounce metrics and ship-rate metrics are indicative of immediately actionnable issues.

----

TL;DR

@knadh  I don't mind going the extra mile and doing the UI work for toggles, but if you want to avoid this PR becoming a :rabbit: :hole: relegating the toggles to a separate PR could make a lot of sense.

knadh:
> @mr-karan @avanier if the API is behind the BasicAuth, then it can just be on all the time, maybe.
https://github.com/knadh/listmonk/pull/694#issuecomment-1060633105

> Yep, it'll be on all the time as discussed ^ .
https://github.com/knadh/listmonk/pull/694#issuecomment-1060798604

Making some toggle-able and others not isn't consistent. It's okay to have to metrics on all the time. I'm guessing that internally it's quite very cheap to just do `Incr()` in a bunch of places. Maybe we could do a benchmark and then decide on this?

mr-karan:
+1 It makes sense for Listmonk to always expose metrics out of the box, as a good default.

The current config looks like:

```
# Prometheus Metrics.
[metrics]
export_process_metrics = true
export_http_metrics = true
```

Both of these, don't really need to be "toggled", as a sane default both of them can always be true.

> At-large HTTP metrics are usually the bread and butter of Prometheus queries, but they run the risk of being high cardinality depending on the size of your deployment. (Also what constitutes high cardinality depends on which Prometheus backend you use, so YMMV.) Either way, they're generally high-ish value.

We took care of the cardinality issue by ensuring the labels aren't dynamic in any metric. Don't think this will be a problem in any listmonk deployment.

> Maybe we could do a benchmark and then decide on this?

+1 @avanier we can probably run a HTTP Load Tests on some API endpoints. In all of the API endpoints, we have a middleware which does .Increment/.UpdateDuration and we can compare the same without metrics.

knadh:
Any updates on this?

avanier:
I was under the impression that there was a desire to do performance testing on this. Otherwise, going over what we discussed, would I be correct in assuming the only changes left is to remove any configuration toggles? This I could probably get in by end of week.

sbusso:
this looks like an almost done PR and interesting feature, is there any chance to get it merged at some point?

knadh:
Yep, this is on my to-do to review and merge in the next release.

avanier:
Sorry for the late answer. If I remember correctly I had limited the path depth collected for those metrics to avoid this cardinality exposition. It's also not the most important thing to monitor. It should be fine to drop the metric and revisit this later if needed.

_(The only legitimate use-case I can think of right now is catching errors on missing assets or the like. But that's something that could just as easily be done at the level of an ingress controller or a CDN.)_

ChrisTG742:
Does this implementation put additional load on your server even if one doesn't use Prometheus at all? Or is it just some sort of REST endpoint that only gets active when called via the API?
I'm asking since there are other tools out there that that have implemented rather cpu-hungry and sluggish Prometheus support and I'm afraid listmonk could lose it's lightweight, too.

knadh:
@ChrisTG742 apps event the Prometheus protocol (simple string formatting) which any system that can read the protocol can consume. It doesn't have to be Prometheus. From a load perspective, there wouldn't be an issue. But what we were discussing was whether fine grained per-URI metrics is actually useful in an app like listmonk which only has a tiny handful of public handlers.

> (The only legitimate use-case I can think of right now is catching errors on missing assets or the like. But that's something that could just as easily be done at the level of an ingress controller or a CDN.)

@avanier I agree. Let's close this PR and revisit metrics support in the future when there's a stronger usecase. Thanks!

PR #707, opened 2022-02-15
Improve the processing of bounced emails

stevesavanna:
- Support multipart format when parsing the header of the bounced message. This is useful when the mail comes through multiple SMTP servers.
- If any of the identifiers are not available in the parsed header, try to extract them from the message body (not just the CampUUID and SubUUID). This is useful when the header is broken for example because of an unexpected new line.
- Fixing issue: before reading the entire body of the bounced message need to reset the pointer of the buffer.


knadh:
Thanks @stevesavanna. Will review this over the weekend.

> so I need to reset the pointer first in order to get the full body (b, _ = c.RetrRaw ( id)).

Was trying to avoid downloading entire message bodies in case there be large attachments or parts that aren't necessary for header extraction.

toontoet:
Using Postfix / docker-mailserver as the SMTP server will always result in bounce-mails with the necessary headers in a mime-part. Therefore bounce handling is not working now using Postfix as the upstream mailserver.
When using the `bounce_notice_recipient` directive in the postfix configuration you'll only get the headers of the original message in a mime-part instead of the whole body. There won't be any large attachments then. 

Having this functionality merged would be awesome and will result in getting the bounce handling to work using Postfix. But maybe a configuration option ('Parse mime-parts yes/no') in the bounce settings would make it more versatile.

knadh:
Hi @stevesavanna. Sorry it took a while for me to get around to reviewing this. Could you please attach a sample multi-part raw body that I can test the changes with?

stevesavanna:
[sample_bounce_email.txt](https://github.com/knadh/listmonk/files/8323456/sample_bounce_email.txt)
[sample_bounce_email_2.txt](https://github.com/knadh/listmonk/files/8323458/sample_bounce_email_2.txt)

Issue #785, opened 2022-04-24
How to escape '<' in email subject?


rohanrajpal:
If I add `<` in email subject

![image](https://user-images.githubusercontent.com/7023147/164971864-37a31157-6a2d-47c7-b7b1-97e9040f8386.png)

it appears as 
![image](https://user-images.githubusercontent.com/7023147/164971887-38ebe055-07c2-4b61-9b58-0e6e42740ce4.png)

How do I escape it?

rohanrajpal:
potentially related issue
https://stackoverflow.com/a/48527166

knadh:
Whoops, this is a bug. The subject line template should use `text/plain`. For now, you can use the `Safe` escape function.

```go
{{ .Subscriber.Name }} {{ Safe "<>" }} Hello
```

rohanrajpal:
> ```go
> {{ Safe "<>" }}
> ```

This works! Thanks, was the above stack overflow link useful or the issue is different?

knadh:
Same issue. Subject line should use plain text template processing instead of HTML.

wweverma1:
@knadh Can I take up this issue?

knadh:
Yes, sure. Just have to change this to `template/text` here: https://github.com/knadh/listmonk/blob/75190d98543c71ca8a827f84aa6ae000aa3cd44e/models/models.go#L439

wweverma1:
@knadh Hi, I've raised a PR against this.
https://github.com/knadh/listmonk/pull/799

PR #799, opened 2022-05-02
Processing subject line as text instead of HTML

wweverma1:
Fixes the issue https://github.com/knadh/listmonk/issues/785


knadh:
What you've submitted is not valid Go code @wweverma1.

I've fixed the issue in https://github.com/knadh/listmonk/commit/89eca5f14bdc145e2baca994261de67434628bdf

Issue #822, opened 2022-06-01
Ability to run app with a one liner at any time


CPLepage:
**Is your feature request related to a problem? Please describe.**
Trying to package listmonk with other services in a docker-compose, but the need to start the db, then the app with --install, then `docker-compose up -d` make everything more complicated. I played around with `--install`, `--idempotent` and `--yes` which brought me to ok results, but a one liner command that install/upgrade if needed and then runs the app through would make it much more easier!

**Describe the solution you'd like**

```yml
  db:
    image: postgres:13
    environment:
      - POSTGRES_PASSWORD=listmonk
      - POSTGRES_USER=listmonk
      - POSTGRES_DB=listmonk
    restart: unless-stopped
    healthcheck:
      test: [ "CMD-SHELL", "pg_isready -U listmonk" ]
      interval: 10s
      timeout: 5s
      retries: 6
    volumes:
      - type: volume
        source: listmonk-data
        target: /var/lib/postgresql/data

  app:
    restart: unless-stopped
    image: listmonk/listmonk:latest
    command: ["./listmonk", "--install", "--idempotent", "--yes", "--run-through"]
    ports:
      - "9000:9000"
    environment:
      - TZ=Etc/UTC
    depends_on:
      - db
    volumes:
      - ./config.toml:/listmonk/config.toml
```

If something like this could fresh install in a new environment or simply start if called on `docker-compose restart`, it would be amazing!

CPLepage:
**UPDATE**

use it like this : 
```yml
  app:
    restart: unless-stopped
    image: listmonk/listmonk:latest
    command: /bin/sh -c "./listmonk --install --idempotent --yes && ./listmonk --upgrade --yes && ./listmonk"
    ports:
      - "9000:9000"
    environment:
      - TZ=Etc/UTC
    depends_on:
      - db
    volumes:
      - ./config.toml:/listmonk/config.toml
```

PR #823, opened 2022-06-01
Issue 822 : run-through flag

CPLepage:
Related to [!822](https://github.com/knadh/listmonk/issues/822)

- added `run-through` to make the app run after the install/upgrade process


knadh:
Hi @CPLepage. install and upgrade are clear, granular commands. `--install` in theory will only ever be run once and `--upgrade`, sparingly. A one liner that combines these into running brings marginal benefits, if any at all.

One can always do `./listmonk --upgrade --yes && ./listmonk`

The usecase you describe in #822 is to have a one line docker-compose command. That can be achieved without having to introduce this new mode in listmonk's core.

CPLepage:
Hi @knadh, damn I never though about the `&&` operator. Works flawlessly as expected! Thanks a lot!
Great job on listmonk btw I really love the app 🙏🙌

Issue #544, opened 2021-10-27
arm64 arch CPU support in binary and docker


GeniusWiki:
Whatever AWS graviton2 or new apple M1,  ARM  CPU becomes more popular.   It is great if the docker can provide ARM64 compatible version, same in binary (although I don't need that).

knadh:
@mr-karan would you like to take a look at this?

alf1e:
I'm currently working on porting this to ARM and in the testing phase as we speak

alf1e:
I am currently stumped on this error
```
 > [2/6] RUN apk --no-cache add ca-certificates tzdata:
#7 0.224 exec /bin/sh: exec format error
```
I am working on a fix though, just very slowly cause this should work

alf1e:
Ah great. The packages `ca-certificates, tzdata` don't support ARM. Is there any workaround you know for this?

knadh:
I'm not familiar with Alpine, unfortunately. @mr-karan?

nemunaire:
FYI, I run listmonk on arm64 hardware since a few months using a Docker container.

I made a multi-stage `Dockerfile` in order to embed all the build steps inside: https://github.com/nemunaire/listmonk/commit/a8c7c2a5d92091883ab4fef5ac888812175d6431
I run the `Dockerfile` to build the image directly on the host (where there is no `npm`, `make`, `go`, ...). I didn't make a PR as I think this is not the way you want to do that.

I can't help as I didn't know goreleaser, but most probably, the error:

    #7 0.224 exec /bin/sh: exec format error

is because the docker image architecture is not the same as the hardware architecture who run the build (eg. image is arm64 and hardware is amd64). Without `qemu` it's not possible to do any `RUN` if architectures are not the same.

(With my *all-in-one Dockerfile*, the common solution is to use `docker buildx` with github actions this way: https://github.com/ddvk/rmfakecloud/blob/master/.github/workflows/release.yml#L50)

mr-karan:
Yeah, buildx is also one way to solve this.

With goreleaser, however it's possible to also use `docker manifest` and create a multi-arch image: https://goreleaser.com/customization/docker_manifest/. Manifest is basically some meta information about the image (and in this context, OS and Arch is what we're interested in). This also uses `buildx` internally to build the image.

Regarding `ca-certificates`, can you try this fix? https://stackoverflow.com/a/70771488/709452 @alf1e 

[...]

PR #849, opened 2022-06-25
Added arm64 arch & changed windows release to zip

arghyadipchak:
- Added arm64 arch support (#544)
- Changed windows release to .zip (instead of .tar.gz) which is easier for windows users

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,6 +19,9 @@ jobs:
         with:
           go-version: 1.17
 
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+
       - name: Login to Docker Registry
         uses: docker/login-action@v1
         with:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,20 +10,20 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
           fetch-depth: 0
 
       - name: Set up Go
-        uses: actions/setup-go@v2
+        uses: actions/setup-go@v3
         with:
-          go-version: 1.17
+          go-version: "1.17"
 
       - name: Set up QEMU
-        uses: docker/setup-qemu-action@v1
+        uses: docker/setup-qemu-action@v2
 
       - name: Login to Docker Registry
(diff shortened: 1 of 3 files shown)


knadh:
Thanks for the PR. Have you tested whether the same tag (`listmonk:latest`) works for multiple architectures on DockerHub?

arghyadipchak:
> Thanks for the PR. Have you tested whether the same tag (`listmonk:latest`) works for multiple architectures on DockerHub?

Made some updates and tested it, everything works fine now.

Updates:
- Builds docker images for both amd64 and arm64
- Windows binaries are released as zip
- Fixed yarn.lock written in .gitignore
- Version upgraded for all github actions

*Docker images are also released with tags {version}-amd64 and {version}-arm64. There is no way to configure goreleaser to solve this problem

mr-karan:
> *Docker images are also released with tags {version}-amd64 and {version}-arm64. There is no way to configure goreleaser to solve this problem

@arghyadipchak `goreleaser` and `docker` CLI both support this. 

I guess there are 2 approaches here: 

1. Using [`docker_manifest`](https://docs.docker.com/engine/reference/commandline/manifest/), we can specify the platform/arch metadata to a list of images. The idea is that we build and push these 2 images separately, but using `docker manifest create` we can combine the image in a single tag. 

Goreleaser does support the use of  [docker_manifest](https://goreleaser.com/customization/docker_manifest/). 

2. There's an even simpler way using `docker buildx`:

Using [buildx]( https://docs.docker.com/desktop/multi-arch/), we can build and give a `--platform` flag as a comma separated list of platforms that the docker image should support. This way we don't have to manually create a manifest.

`docker buildx build --platform linux/amd64,linux/arm64/v8 --tag listmonk:latest`

^ This should produce an image that will work on both the architectures.

So, basically if we want to push explicit platform tags then we need to use `docker manifest`. If we want a single tag that should work irrespective of the platform, we can simply add a platform flag.

knadh:
> If we want a single tag that should work irrespective of the platform, we can simply add a platform flag.

Thanks @mr-karan. This is the correct approach.

knadh:
@arghyadipchak would you like to amend the PR?

arghyadipchak:
> @arghyadipchak would you like to amend the PR?

I would love to, but currently I don't have time. I am closing this PR for now. I will find a better approach and contribute a new PR as soon as I can.

PR #853, opened 2022-06-26
Add Import system By copy pasting for noob

BCW52:
For benginner it very hard to add subscriber using csv file


knadh:
You've incorrectly opened a PR instead of opening an issue.

To copy-paste and import, you'll have to format records in a specific certain way (email, name, attribs) and prepare many lines, which you'll eventually copy-paste. If you just save the same data to a .csv file and upload, that's it, right?

PR #856, opened 2022-06-28
feature: Make config file optional

sbusso:
Config file should be optional as we can load all configs from env vars and we should validated required settings are defined.


knadh:
This'll start listmonk without a username and password, which is dangerous. The empty username and password in config.toml is not default behaviour. `--new-config` generates a random password always. The user has to explicitly clear the username and password by the editing the config file to get an unauthenticated instance. This behaviour only exists for backwards compatibility and will go away in the future.

PR #869, opened 2022-07-14
fix(i18n): update i18n/jp for 2.2.0 release

t3hmrman:
This commit contains translations for the new terms entered before release 2.2.0


t3hmrman:
OK this should be ready to go now, just a couple notes on what was wrong for others:

1. New translations were added so those have been translated
2. Language name should have been "Japanese" (日本語) but was English (in Japanese)
3. English replacements were translated (ex `{num}`) 

Both 2 & 3 were mistakes that I should have noticed earlier :bow:

Issue #851, opened 2022-06-25
Next release (v2.2.0)


laupolo6:
Hi,
When will the next major/minor version be released? any guess?
I think it would be great if listmonk have a more predictable timeframe or roadmap for future releases.
thanks

knadh:
Please see this for some additional context on release cycles and plans: https://github.com/knadh/listmonk/issues/675

That said, the next release is already at least a month behind the schedule I had in mind. Since I happen to be the only developer working on the project for now, this unfortunately entirely depends on the limited personal time I can dedicate.

v2.1.0 is a good-enough, feature complete, stable version that's in large production environments, so really, I haven't been scrambling to make a new release either. The current big WIP is transactional e-mailing and once that is done, the next version should be release. I'm hoping this will happen in the next ~2 weeks.

sbusso:
@knadh first thank for the amazing work on listmonk. I have managed my own setup for more than 10 years, but now it is hard to maintain and listmonk is the closest to what I was doing. Is there any public roadmap or list of features you are looking at improving / implementing?

knadh:
Thanks @sbusso. The GitHub issues are properly tagged. The ones I've picked to work on are assigned to me.

All the other open issues tagged `enhancement` are good to be discussed and worked on.

knadh:
[Transactional mailing](https://github.com/knadh/listmonk/commit/463e92d1e1353a79a8fd8d32c75fcef5e38dee5a) is finally merged. Had to wait for a week to test the changes in production at scale. Now, a handful of pending tweaks and minor bug fixes and the release is good to go. The coming weekend :crossed_fingers:.

NicoHood:
Can we see a fix for https://github.com/knadh/listmonk/issues/712 for the next release?

knadh:
Tagging i18n language contributors. There are a few new strings in the English file that aren't in the other files. Would be great if you'd be able to find time to compare and submit any changes before the next release. Thank you!

- čeština (@srchlm)
- Deutsch (@jonathanmmm @an0nfunc @spezifisch @dunklesToast @ChrisTG742 @m42e @NicoHood)
- Español (@jorge-vitrubio @dreglad @seba81)
- Français (@nathanaelhoun @jonathandhn @MickGe @FelixDz @jonathandhn)
- Hungarian (@bohemtucsok)
- Italiano (@etcware @jorge-vitrubio @marcofucito @TomBoss)
- 英語 (@t3hmrman)
[...]

PR #873, opened 2022-07-14
Update de.json

ChrisTG742:
See https://github.com/knadh/listmonk/issues/851


i18n/de.json
```diff
@@ -454,10 +454,10 @@
     "settings.smtp.name": "SMTP",
     "settings.smtp.retries": "Wiederholungen",
     "settings.smtp.retriesHelp": "Maximale Anzahl an Wiederholungen, wenn eine Machricht fehlschlägt.",
-    "settings.smtp.sendTest": "Send e-mail",
+    "settings.smtp.sendTest": "E-mail senden",
     "settings.smtp.setCustomHeaders": "Benutzerdefinierten Header verwenden",
-    "settings.smtp.testConnection": "Test connection",
-    "settings.smtp.toEmail": "To e-mail",
+    "settings.smtp.testConnection": "Verbindung testen",
+    "settings.smtp.toEmail": "An E-mail",
```
an0nfunc, line 460: Isn't this more like "Empänger E-Mail" than "An E-Mail". That sounds strange. Do you use it like that in normal every-day speech?

ChrisTG742: The question is the context this is used in. If it is ment like "recipient", then I'm with you that should translate to "Empfänger". But I thought is is ment as in the setence "should be sent **to**", then "**An**" fits better here.
Maybe @knadh could shed some light in here?
TIA

knadh: image

This is how it looks on the UI. Sending test message `from email` -> `to e-mail`.

ChrisTG742: Okay, then @an0nfunc is right, I'll fix that.

PR #878, opened 2022-07-14
update FI translation (still partial)

pucilpet:


knadh:
Thanks! Looks like the indentation has changed from the original. Could you please change the indentation back to spaces instead of tabs?

pucilpet:
> Thanks! Looks like the indentation has changed from the original. Could you please change the indentation back to spaces instead of tabs?

Done.

PR #903, opened 2022-08-12
fixes missing list items on opt-in confirmation page

razorness:
additional changes:
- Extend .gitignore for windows and JetBrains IDE.
- Fixed typo in german translation.
- Dont force to doctype to be `<!doctype html>`.
- Added template function `safe` to escape conditional comments like `<!--[if mso]>`. Better solution would be to use `html/template` instead of `template/template` .


cmd/init.go
```diff
@@ -586,6 +586,9 @@ func initNotifTemplates(path string, fs stuffbin.FileSystem, i *i18n.I18n, cs *c
 		"L": func() *i18n.I18n {
 			return i
 		},
+		"safe": func(s string) template.HTML {
```
knadh, line 589: Let's rename this to `Safe` (uppercase S) to be consistent with the other exported functions.

PR #922, opened 2022-08-27
Translated the rest of the lines

organicnz:


organicnz:
Translated the rest of the lines

PR #955, opened 2022-09-22
Create ca.json

davidesteve:
This adds Catalan language support to listmonk.


ChrisTG742:
If you want your merge request to be respected, you should at least write a few lines about what it is for or what it does.

AFAIC this adds Catalan language support to listmonk, right?

PR #990, opened 2022-10-20
Allow also non-subscribers to receive transactional emails

weese:
Currently transactional emails are rejected for subscriber_emails that are not already stored in the subscribers table. However, the point of transaction emails is to for example send a verification email at account creation before an email is confirmed. In these cases we don't subscribe a user to mailing list but might send that email exactly once.

Long story short, that PR allows api/tx to accept subscriber_emails that are not known already. I've successfully tested it locally.

### DIFF
--- a/cmd/tx.go
+++ b/cmd/tx.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"errors"
 	"fmt"
 	"net/http"
 	"net/textproto"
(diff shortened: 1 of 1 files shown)


knadh:
This patch breaks the default expected behaviour though. The `tx` API accepts either the `subscriber_email` or `id`. If an `id` is sent and it doesn't exist, this check breaks.

The better way of doing this would be to make it configurable.
- When `email` is sent, if the subscriber doesn't exist, send anyway?
- If the subscriber doesn't exist, create the subscriber?

weese:
Hey, the change will only change the behaviour when the API is used with a non-empty `subscriber_email`. If a `subscriber_id` is given, `subscriber_email` is empty and the the inner if condition is fulfilled and exits if the subscriber id is unknown.

weese:
Configurable would be nice, but I couldn't think of a use case where I would use concrete email addresses and want the api call to fail if it is unknown.

knadh:
> I couldn't think of a use case where I would use concrete email addresses and want the api call to fail if it is unknown.

This is quite common. A lot of customer databases uses email as the internal primary key.

> In my use case I don't want to persist email addresses as subscribers in Listmonk due to GDPR requirements. We need to keep the stored personal data minimal.

Understood. However, since the API's contract right now is to fail on unknown e-mails and IDs, we should only introduce this change via an additional parameter. Something like `accept_non_subscriber = true` that enables this behaviour.

charlesmudy:
Very nice feature. I would suggest thought to make it optional just as @knadh suggested since we might want to track view and click which I believe is possible if the subscriber is recorded right?

Also, will this apply to other messenger such as SMS?

PR #993, opened 2022-10-21
Fix bug in template cache that stored only references of the last template

weese:
The template cache `tpls` stores pointers to templates and is initially loaded with the existing templates using a reference to a for-loop variable. As the variable is the same and just the value changes, the reference stays the same, so that all entries in `tpls` are identical with the consequence that the whole cache stores just the last loaded template.

I fixed it by creating a copying of the loop-variable and store the reference to that one. Ideally, you could consider storing the templates by-value instead of by-reference in `tpls` to prevent these little issues 😊 

Another thing I'm worried about: The in-memory template cache prevents Listmonk from being horizontally scaled, what you would typically do in production in k8s. Mutliple Listmonk replicas in a cluster may start with identical caches, but get out of sync the moment a template is created/updated as that is processed. Only the one replica processing the request has the lattest cache entry.

My proposal is to either:
1. completely remove the cache
2. allow to disable the cache via config
3. give the cache entries a short TTL

What do you think?

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -443,7 +443,8 @@ func initTxTemplates(m *manager.Manager, app *App) {
 			lo.Printf("error compiling transactional template %d: %v", t.ID, err)
 			continue
 		}
-		m.CacheTpl(t.ID, &t)
+		var copy models.Template = t
+		m.CacheTpl(t.ID, &copy)
 	}
 }
 
--- a/cmd/tx.go
+++ b/cmd/tx.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"errors"
 	"fmt"
 	"net/http"
 	"net/textproto"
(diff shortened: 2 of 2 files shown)


knadh:
Darn! The notorious Go for loop, reference pitfall! I fixed it here: https://github.com/knadh/listmonk/commit/281c47198cf31ea004af44dff4989413fe26dd04

Also, just realised that this PR has two separate things.

Completely removing the cache isn't ideal, and it's not ideal for such a low level behaviour to be configurable on the UI. Without caching, every transaction will pull the template from the DB, parse and compile it into a Go template, which is extremely slow and wasteful, especially when the volume of transactional e-mails grow.

Horizontally scaling listmonk isn't possible as campaign management is heavily dependent on the database state and internal runtime state.

weese:
> Also, just realised that this PR has two separate things.

Oops, that was not itentional. I didn't want to sneak in my other code change 😊

knadh:
> Then maybe a TTL for cache entries might be an option. Oh, just for me to understand. Besides the cache what other state is in the pod that is not persisted in the DB?

A TTL won't solve the horizontal scaling problem. Before the TTL hits, you can still have stale templates.

> Besides the cache what other state is in the pod that is not persisted in the DB?

The entire campaign management state when a campaign is running is in-memory. If two instances run at the same time, campaigns will end up sending duplicate e-mails.

PR #1013, opened 2022-11-08
Added support to run list monk within IIS

kenrf:
This is a modification to allow listmonk to run within IIS. 
**Problem:**
By default IIS assumes your app will be able to be told what random port to run on at each IIS/webapp startup. Listmonk had no way of knowing what port it should launch on (IIS sets it and it's up to the web app to find and use it).
**Solution:**
This change allows listmonk to check for the ASPNETCORE_PORT as set by ACNM, and then use this value for the listening port, and if no value is found, use the config.toml value as per usual.
There's also the addition of a web.config file that allows IIS to run listmonk as an aspnetcore web app. It's generic and simply tells IIS what to launch.
**Notes:**
These settings changes are specific to windows builds (.exe)
Hopefully these help someone run under IIS and don't break anything else.

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -707,18 +707,25 @@ func initHTTPServer(app *App) *echo.Echo {
 	}
 
 	// Register all HTTP handlers.
-	initHTTPHandlers(srv, app)
-
-	// Start the server.
-	go func() {
-		if err := srv.Start(ko.String("app.address")); err != nil {
-			if strings.Contains(err.Error(), "Server closed") {
-				lo.Println("HTTP server shut down")
-			} else {
-				lo.Fatalf("error starting HTTP server: %v", err)
-			}
-		}
-	}()
+    initHTTPHandlers(srv, app)
+    port := ko.String("app.port")
+    if os.Getenv("ASPNETCORE_PORT") != "" { // get enviroment variable that set by ACNM
+        port = os.Getenv("ASPNETCORE_PORT")
+        lo.Println("port retrieved from aspnetcore")
+    } else {
+        lo.Println("Could not find aspnet core port via IIS")
+    }
+    // Start the server.
+    lo.Println("HTTP starting on: " + "app.address" + ":" + port)
+    go func() {
+        if err := srv.Start(ko.String("app.address") + ":" + port); err != nil {
+            if strings.Contains(err.Error(), "Server closed") {
+                lo.Println("HTTP server shut down")
+            } else {
+                lo.Fatalf("error starting HTTP server: %v", err)
+            }
+        }
+    }()
 
 	return srv
 }
--- a/web.config
+++ b/web.config
@@ -1,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+    <system.webServer>
+        <handlers>
+            <add name="aspNetCore" path="*" verb="*" modules="AspNetCoreModule" resourceType="Unspecified" />
+        </handlers>
+        <aspNetCore processPath=".\listmonk.exe" />
+    </system.webServer>
+</configuration>


knadh:
Wouldn't it be possible to pass custom environment variables in the IIS environment? `LISTMONK_app__address` variable works to pass the port config.

https://listmonk.app/docs/configuration/#environment-variables

kenrf:
I don't believe it is :-( more info here, infact I think I used this example to solve the hosting a golang app in IIS issue I had so I could run list monk:
https://medium.com/@mossila/running-go-behind-iis-ce1a610116df
The problem is that the internal IIS port your app is expected to be listening on is dynamic, and figured out by IIS at run time. 
I could never figure out another way around this, (including forcing IIS to use a fixed 'internal' port).

knadh:
It's not ideal to add vendor specific logic into the app's core.

Rather than hardcoding this logic, wouldn't it be possible to write a startup script (`.bat` file on Windows) that received the `ASPNETCORE_PORT` env var and passed it to listmonk? Then, you get IIS to execute the script instead of listmonk directly.

Something like `listmonk.bat` (pseudocode)

```bat
:: Get the ASPNETCORE_PORT variable from IIS and copy it to LISTMONK_app_address
SET LISTMONK_app_address="0.0.0.0:%ASPNETCORE_PORT%"

:: Execute listmonk
listmonk.exe
```

kenrf:
From what I have read, and tried, unfortunately a .bat file wont work either. This is one of the "features" of running a non .net language in IIS I guess. Totally get the need to not pollute code with vendor specific code, however I've found no other way around this.
If done elegantly and safely I think it's ok, given the potential large number of IIS users out there, maybe of which a number have tried to run listmonk in IIS already and failed.
This is a tough one!
Happy to try any other work arounds you can think of in my dev/test env too.

   Ken.

knadh:
https://stackoverflow.com/questions/415409/run-batch-file-as-a-windows-service

Does this make sense?

kenrf:
Unlikely from what I have read, the problem we’re trying to solve is “iis passes a dynamic port that the container app is expected to acknowledge and use” this port changes on every restart. That link seems more related to actually starting an app. Who else do we know that is running listmonk from within iis? Would be interesting to see how they’re doing it. Thanks,   KenSent from my iPhoneOn 20 Nov 2022, at 5:03 pm, Kailash Nadh ***@***.***> wrote:﻿
https://stackoverflow.com/questions/415409/run-batch-file-as-a-windows-service
Does this make sense?

knadh:
@kenrf the link is about running a bat file as a service. If that's possible, then the earlier discussed approach can be applied. .bat file receives the `ASPNETCORE_PORT` port and passes it to listmonk while starting it as `LISTMONK_app_address`. The bat file acts like a bootstrap starter for the listmonk binary.

kenrf:
Have never heard of a bat file working like that in iis, however that has me thinking, maybe I write a simple exe that can do the same thing, then this exe can be included somewhere in the repo in case people need to deploy to iis, will test this theory out over the coming days.Crossing fingers!   KenSent from my iPhoneOn 20 Nov 2022, at 6:59 pm, Kailash Nadh ***@***.***> wrote:﻿
@kenrf the link is about running a bat file as a service. If that's possible, then the earlier discussed approach can be applied. .bat file receives the ASPNETCORE_PORT port and passes it to listmonk while starting it as LISTMONK_app_address. The bat file acts like a bootstrap starter for the listmonk binary.

kenrf:
This was the original hint that lead me to the go code based solution too:Running Go behind IISmedium.com  KenSent from my iPhoneOn 20 Nov 2022, at 7:08 pm, ken fretwell ***@***.***> wrote:﻿Have never heard of a bat file working like that in iis, however that has me thinking, maybe I write a simple exe that can do the same thing, then this exe can be included somewhere in the repo in case people need to deploy to iis, will test this theory out over the coming days.Crossing fingers!   KenSent from my iPhoneOn 20 Nov 2022, at 6:59 pm, Kailash Nadh ***@***.***> wrote:﻿
@kenrf the link is about running a bat file as a service. If that's possible, then the earlier discussed approach can be applied. .bat file receives the ASPNETCORE_PORT port and passes it to listmonk while starting it as LISTMONK_app_address. The bat file acts like a bootstrap starter for the listmonk binary.

knadh:
Yep, a runner-EXE would work. If it's a .bat file, it's easier to maintain and distribute, that's all. Closing this issue. Once you have something working, we can include that in the official docs.

Issue #1016, opened 2022-11-08
Add support for serverless Postgress


unownone:
**Cannot Connect from Serverless Database Solutions**
The current Implementation doesn't work if you are trying to use a serverless postgres instance for using as a production database. It breaks because serverless databases during connection require a options url parameters like `?options=` with a value `--cluster=cluster_name` . This breaks in the current implementation, hence serverless instances from services like cockroachdb does not work.

**Proposed Solution**
Accepting an extra options parameter which may or may not be provided in order to connect to databases that are hosted on a multi tenant system.

PR #1017, opened 2022-11-08
Issue 1017: Connection to Serverless Database Clusters

unownone:
[See Issue : Issue-1016](https://github.com/knadh/listmonk/issues/1016#issue-1440776166) 
# Summary 
This pull request , attempts to fix the issue while connecting to serverless PostgreSQL instances like cockroachdb, etc. It does so by passing an extra argument `options` which is supported by `sqlX` and subsequently `databases/sql` of Golang . 

# Issue
Migrations cannot be run as cockroachdb doesn't support DROP TYPE CASCADE . Which inturn is making everything miserable.
# Attempted Fixes:
- Running migrations from the schema.sql on Cockroachdb's online migration manager - Didn't Work ❌
- Running migrations from cockroach CLI from schema.sql - Didn't Work ❌
- Running migrations from cockroach CLI from shema.sql while removing CASCADE from all DROPs - Partially Worked ✅

Upon this if we try to run it runs into a migration issue , and migration then causes a bunch of other issues while trying to apply them!

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -248,6 +248,7 @@ func initDB() *sqlx.DB {
 		MaxOpen     int           `koanf:"max_open"`
 		MaxIdle     int           `koanf:"max_idle"`
 		MaxLifetime time.Duration `koanf:"max_lifetime"`
+		Options     string        `koanf:"options"`
 	}
 	if err := ko.Unmarshal("db", &c); err != nil {
 		lo.Fatalf("error loading db config: %v", err)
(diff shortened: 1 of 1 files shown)


knadh:
Thanks @unownone. The PR description doesn't describe the actual change in the PR. Could you please explain?

unownone:
@knadh Surely. The change is addition of a `Options` struct attribute which passes down the `options` attribute down to the sqlx handler.  The Options attribute is added in the c struct that is defined in the initDB() method . While initialization we can also pass down a options ( for serverless or other databses ) by setting a `options` option under [db] in config.toml or setting a env variable `LISTMONK_db__options` , it accepts a string and uses it to build the connection URI.

Example : 
If you are using cockroachdb, it provides you a serverless instance for free. To connect to it you have to specify other than the db name host port and password , also another url parameter called options which provides the cluster name . It looks somewhat like this : 
` postgres://<USERNAME>:<PASSWORD>@<host-uri>/<db-name>?sslmode=<ssl>&options=<cluster option>`

knadh:
Makes sense. Thanks.

knadh:
Is this ready to merge?

unownone:
@knadh let me do some testing and will create a pr!

knadh:
https://github.com/knadh/listmonk/commit/5d4f1ea0ad8cbcfac0285659d6e60fd2dec4064c addresses this.

PR #1021, opened 2022-11-12
Cosmetic. Centering the logo in the header

xmacaba:
Cosmetic changes. Looks way better when the logo is centered.


knadh:
Thanks @xmacaba, but that's subjective! I prefer the default listmonk logo on the left. Depends on the type of logo really.

You can easily customize this on your instance by adding a line of CSS in Settings -> Appearance. No need to change the core.

PR #1056, opened 2022-11-24
Partial update of NL i18n

TychoWerner:


i18n/nl.json
```diff
@@ -14,10 +14,10 @@
     "bounces.unknownService": "Onbekende service.",
     "bounces.view": "Zie bounces",
     "campaigns.addAltText": "Voeg alternatieve tekst zonder opmaak toe",
-    "campaigns.archive": "Archive",
-    "campaigns.archiveEnable": "Publish to public archive",
-    "campaigns.archiveHelp": "Publish (running, paused, finished) the campaign message on the public archive.",
-    "campaigns.archiveMeta": "Campaign metadata",
+    "campaigns.archive": "Archiveren",
+    "campaigns.archiveEnable": "Publiceren naar publiek archief.",
```
Jjagg, line 18: There's no trailing period in the original English text here.

```suggestion
    "campaigns.archiveEnable": "Publiceren naar publiek archief",
```

i18n/nl.json
```diff
@@ -14,10 +14,10 @@
     "bounces.unknownService": "Onbekende service.",
     "bounces.view": "Zie bounces",
     "campaigns.addAltText": "Voeg alternatieve tekst zonder opmaak toe",
-    "campaigns.archive": "Archive",
-    "campaigns.archiveEnable": "Publish to public archive",
-    "campaigns.archiveHelp": "Publish (running, paused, finished) the campaign message on the public archive.",
-    "campaigns.archiveMeta": "Campaign metadata",
+    "campaigns.archive": "Archiveren",
```
Jjagg, line 17: Sorry for the separate review, but I think this should be the noun archive ('Archief'), not the verb ('Archiveren'). At least that's how it is in the german translations. @knadh?

knadh: Yep, it should be the noun.

knadh: @TychoWerner Could you please amend the PR with this change?

TychoWerner: These changes are valid.

PR #1059, opened 2022-11-25
Update ca.json

davidesteve:
Update Catalan version for v2.3.0 #1039


davidesteve:
Hope it works like that.

PR #1112, opened 2022-12-19
Proposal: add per-provider rate limit

### LINKED ISSUE #1097: Cold Email Server Warm Up Request
First of all, Thank You for an excellent product.

Would it be possible to enhance the performance section of settings to include a percentage increase in the sliding window?

I'd like to be able to use ListMonk to warm up a cold email server automatically rather than manually. For instance, using a subscriber list of 10,000 users, starting at 200 users and increasing by 10-50% per 24hr period until all 10,000 are sent the campaign.

i.e. for a campaign with a sliding window of 15% starting at 200 users:
Day 1: 200 / Day 2: 230 / Day 3: 265 / Day 4: 305 / Day 5: 351..... etc until 10000 is reached.

I hope I've explained this correctly.

Colin.

ajoslin:
I use multiple SMTP providers, and we are currently adding a couple. As we warm these providers up, would like to set a daily limit for these providers, so that when the limit is hit it will move to another one.

For example: new SES account at 50,000 per day. 
image

Fixes #1097


ajoslin:
Looking at it, need to clean up the code and add i18n, this is just a POC

The while loop needs an escape hatch as well

knadh:
Hi @ajoslin.

- How does this interact with the global sliding window limit?
- A daily limit isn't universal. If at all, it should be a custom time period. `X` messages per `Y` duration, like the sliding rate limit.

halms:
I think that a per-provider rate limit is a very good enhancement.
I'd say this should be implemented like the sliding rate limit (maybe also allowing 'per day') as the window.

From my point of view, this could replace the global sliding window limit, Maybe I'm missing something, but I'd say the typical use case for the sliding window is to obey provider restrictions. But I'm happy to be convinced otherwise.

What I'd personally need is a different limit for Emails (SMTP) and other messengers (where I'm working to implement a custom backend around `signald` and `whatsmeow`). A limit per provider (SMTP servers and messengers) would fulfill this very nicely.

internal/messenger/email/email.go
```diff
@@ -99,10 +113,17 @@ func (e *Emailer) Push(m messenger.Message) error {
...
-		srv = e.servers[0]
+
+	for {
+		if ln > 1 {
+			srv = e.servers[rand.Intn(ln)]
+			if (srv.limiter.Allow()) {
+				break;
+			}
+		} else {
+			srv = e.servers[0]
```
kamikazechaser, line 124: A limiter token needs to be consumed here when only 1 server is configured with a rate limit.

kamikazechaser (commented):
I think this is a good feature. Possible gotcha of a negligible (but possibly important) performance hit of rate limiting on high throughput campaigns.

Also ,the sliding window in the manager already applies a rate limit so this could be redundant if someone opts to apply a global rate limit.

This could also block the workers once tokens are used up. I can possibly see some side effects of that if there is a mix of rate-limited and unlimited servers.

ajoslin:
> * How does this interact with the global sliding window limit?

This is about each individual provider. For example, right now we have a new plan with a new SMTP provider at a great deal but they require a long 60 day warmup period. Each day I go in and edit the daily quota.

In the meantime, we continue to allow other providers to fill in the remaining messages sent daily.

> * A daily limit isn't universal. If at all, it should be a custom time period. `X` messages per `Y` duration, like the sliding rate limit.

Sure, the duration could be configurable. I like that. I just think of it in days.

---

I created a modified version of this that uses the Redis rate limiter and things still send fast enough to 1M+ subscribers.

I also added an escape hatch when there are none available to throw an error too, which will pause the campaign.

I'll upload it soon to this PR since it seems to be getting good feedback.

knadh:
Closing this as there are unresolved issues and `master` has moved ahead.

PR #1115, opened 2022-12-20
Fix misleading German translation

jfilter:
The word `erfolgreich` (successfully) implies the user is already confirmed, even though s/he still has to click on the confirmation link.


NicoHood:
But well, isnt that a general issue then, also in english?

jfilter:
The English version for `public.confirmSubInfo` is currently:

> "You have been added to the following lists:"

It's not: 

> "You have been added *successfully* to the following lists:".

IMO, this change brings the translation closer to the original.

NicoHood:
alright, agree

PR #1117, opened 2022-12-24
updated spanish translation

jorge-vitrubio:
updates in the translation file for spanish language


jorge-vitrubio:
I will separate languages in branches.

Issue #697, opened 2022-02-13
Adding aria-labels, ids to public forms for accessibility


candideu:
# **Is your feature request related to a problem? Please describe.**
Hello! I've been really enjoying the latest release, especially the ability to edit the css and js of the public pages and admin panel.

That said, I noticed that the public subscription forms are missing key attributes like #ids, `for=`, and aria-labels. These are helpful for accessibility, or to further customize things like placeholders (see: [Labeling Controls](https://www.w3.org/WAI/tutorials/forms/labels/)).

![image](https://user-images.githubusercontent.com/55474996/153737291-5e005b50-40c1-4ad1-b3e5-1083fcd3b361.png)

# **Describe the solution you'd like**
Code at the moment:

```html
  <p>
    <label>E-mail</label>
    <input name="email" required="true" type="email" placeholder="E-mail" autofocus="true">
    <input name="nonce" class="nonce" value="">
  </p>
  <p>
    <label>Name (optional)</label>
    <input name="name" type="text" placeholder="Name (optional)">
  </p>
```

I propose something like this:

```html
  <p>
    <label for="signup-form-email">E-mail</label>
    <input name="email" required="true" type="email" placeholder="E-mail" autofocus="true" id="signup-form-email" aria-label="email">
    <input name="nonce" class="nonce" value="">
  </p>
  <p>
    <label for="signup-form-name">Name (optional)</label>
    <input name="name" type="text" placeholder="Name (optional)" id="signup-form-name" aria-label="name">
  </p>
```

candideu:
Thanks for the fix @shrilakshmishastry

PR #1134, opened 2023-01-06
accessibility fix for subscription form

shrilakshmishastry:
**What's the issue :**
Currently, there are no ids or aria-labels in the public subscription forms. These are helpful for accessibility.

[Link to issue](https://github.com/knadh/listmonk/issues/697)

**What's the solution:**

Add _for_ attribute to label tag and _id_ attribute to input tag. 

Label tag with _for_ attribute takes care of handling accessibility for input tag. Refer this document for more info [doc](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#accessibility_concerns)

**Screenshot of proof:**
image

### DIFF
--- a/static/public/templates/subscription-form.html
+++ b/static/public/templates/subscription-form.html
@@ -6,14 +6,14 @@
     <form method="post" action="" class="form">
         <div>
             <p>
-                <label>{{ L.T "subscribers.email" }}</label>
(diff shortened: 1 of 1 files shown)


candideu:
Thank you @shrilakshmishastry !

Issue #641, opened 2021-12-30
Start of Week on Mondays


NicoHood:
Checkout the schedule calendar:
![grafik](https://user-images.githubusercontent.com/6888294/147766600-d8ee3fd5-1f88-417b-a295-08fbe853a424.png)

The week starts on sunday. In Europa it usually starts on mondays. How can I change this? I think it would make sense to read that information from the browser? Or must I set environment variables for that?

knadh:
There's no config for this. I assumed the [datepicker](https://buefy.org/documentation/datepicker) lib would automatically pick up the locale and adjust accordingly, but that doesn't seem to be the case. Will investigate how this can be done automatically with the lib.

marcinkunert:
I've done some digging and there is a possibility to change the first day of month https://github.com/buefy/buefy/issues/276 (https://github.com/buefy/buefy/commit/59a0be886719d00efebaf5799231af49d8a9fc5f) using `first-day-of-week`` (e.g. the `:first-day-of-week="1" `. 

I don't think creating an option in settings in listmonk is a good idea. I've created a feature request for buefy https://github.com/buefy/buefy/issues/3648 hoping that they will make the start of the week day locale dependent.

knadh:
> I've created a feature request for buefy buefy/buefy#3648 hoping that they will make the start of the week day locale dependent.

Thanks!

MJacred:
The referenced issue got closed with the comment to use the global config https://github.com/buefy/buefy/issues/3648#issuecomment-1021497541 (and that the feature might come someday)

1000i100:
I don't understand, is it fixed or not ? It looks like there is a solution with config file ? (i'm not fully fluent in english)
As of v3.0.0 the week start on sunday and i'vent seen anything to change that nor in config file or in admin UI.

PR #1138, opened 2023-01-08
Fix: Off-by-one error for weekdays in DE i18n

lheimes:
In the German translation, weekdays were off by one.

Might have been caused by the change in bea1680.


knadh:
Related: https://github.com/knadh/listmonk/issues/641

Issue #972, opened 2022-10-07
add go-native e2e test cases for backend


nehagup:
**Is your feature request related to a problem? Please describe.**
Currently, [ integration tests](https://github.com/knadh/listmonk/tree/master/frontend/cypress/integration) are written using Cypress. Quick benefits for go e2e tests.  

-> Reports on Backend(Go) Test Code Coverage.
-> Backend Contributors can write go-native test cases independently without needing to understand Frontend/Cypress.  
-> Opinionated - Backend Tests are less brittle than frontend tests in the long run. 

**Describe the solution you'd like**
I can make a PR which adds go-native e2e tests generated using Keploy which will run as part of the GitHub actions pipeline. It will also take care of the Postgres dependency mocks. With that, we'll also have a test coverage check as part of the GitHub-actions pipeline for the PRs and releases.

knadh:
Thanks @nehagup. Would it be possible see how these tests look, if it's not too much effort? Just trying to understand how it works.

That said, Cypress frontend tests are still necessary because listmonk is primarily a web app with complex (and often brittle with DOM/JS lib weirdness) state management on the UI. It's only mode of operation is via the web app UI.

nehagup:
Sure @knadh I'll make a PR for this in the upcoming week.

nehagup:
@knadh how do you currently mock the SMTP server during testing? Do you run the mock SMTP server or you use some test-credentials?

knadh:
I run a local instance of https://github.com/mailhog/MailHog. It's excellent.

re-Tick:
Hi @knadh, I've a created a PR which show's how Keploy could generate tests for listmonk. I integrated keploy into the codebase (this requires passing context to every dependencies, so made relevant changes) and then used the listmonk UI to record bunch of API tests, as yamls in the tests folder. The current code coverage is 49.2% from these tests -  which can be added by either writing go tests manually or generating more via keploy.

The SQL calls are also recorded as editable yaml stub files corresponding to the API calls, and are saved in the mocks folder. This PR took time because we were in the midst of revamping our SQL driver - to create readable and editable stub. files, and wanted to show you that version.

Looking forward to your thoughts

knadh:
Hi @re-Tick. How do I run the tests locally?

re-Tick:
I have edited the test job in Makefile. So, the steps for running this locally are:
[...]

PR #1140, opened 2023-01-11
test: integrate keploy generating api tests and sql stubs

re-Tick:
ref: https://github.com/knadh/listmonk/issues/972


knadh:
Hi @re-Tick. Whew, that's a big PR! I'll need find some free time to understand how this works and play around with it.

re-Tick:
Sure. In Summary:
1. I am passing request context to SQL methods, so that keploy can generate the correct sql stubs for API tests. 
2. generated tests and sql stubs/mocks using keploy, currently at 49.2% coverage.
3. added a Github CI Action to run the recorded Keploy test cases
4. changed the test of Makefile to run the tests and give code coverage.

knadh:
Hi @re-Tick, while I managed to run this and understand out how Keploy works, I'm holding off on merging this for the next release. The prod build of the `keploy` branch of the listmonk binary is 22MB as opposed to the non-Keploy `master` build which is 13MB. Any thoughts on this?

re-Tick:
Currently, there seems to be some shared models between the Keploy's server and Go-SDK which is leading to importing all the packages of the Keploy server in the final build. 

 I'm working on a fix to ensure that the go compiler doesn't package the entire Keploy server into host application builds.

re-Tick:
Hi @knadh, I've optimised and removed some unnecessary imports and this should optimise the size by around 3mb. After this optimisation the biggest contributor is the gRPC client libs and its used internally by keploy.

knadh:
Hi @re-Tick. The resulting binary is now 21MB instead of 22MB.

Either way, as a general concept, I don't think it's ideal for a testing/dev library to bundle itself (especially at a significant addition to size) into production builds. Maybe a build tag based approach can be explored to stub out the heavy dependencies during build.

re-Tick:
Hi @knadh, Thanks for the feedback. We are releasing a new version of keploy which would not need code changes to the application and generate edge cases automatically. It uses eBPF to hook into OS-level calls rather than language level methods.
I'll create a new PR once we release that :slightly_smiling_face:

PR #1148, opened 2023-01-16
fix: check public URL before presigned URL generation

### LINKED ISSUE #1141: Using private S3 bucket and public CloudFront (CDN) URL together does not work correctly
**Version:**
 - listmonk: v2.3.0
 - OS: Ubuntu 22.04 LTS / Docker

1. Go to the AWS Console and create an S3 bucket with the default (private) settings.
2. Create a new IAM user and give them permissions for the bucket.
3. [Create a CloudFront distribution for the bucket](https://aws.amazon.com/premiumsupport/knowledge-center/cloudfront-https-requests-s3/).
4. [Give the CloudFront distribution access to the bucket](https://aws.amazon.com/premiumsupport/knowledge-center/cloudfront-access-to-amazon-s3/), so it can serve files even though the bucket and its contents are private and cannot be accessed directly.
5. Configure aws cli using the access key and access secret of the new IAM user.
6. Use `aws s3 cp myfile.jpg s3://mybucket/` command to put a file in the bucket. Confirm in the console that the file was uploaded successfully. This also confirms that the IAM credentials are correct and have permission to upload to the bucket.
7. In a web browser visit https://[origin].cloudfront.net/myfile.jpg to verify that the file is being served publicly, even though the s3 bucket is still private.
8. In a web browser visit https://[bucketname].s3.amazonaws.com/myfile.jpg to verify that access to the file is denied as it should be because the bucket is not public. 
9. Go to listmonk settings for media uploads.
10. Set provider to S3 and choose the correct region.
11. Put in the aws access key and aws secret key of the new IAM user.
12. Put in the bucket name of the new bucket.
13. Set bucket path to /
14. Do not edit S3 backend URL. Leave it as the default https://s3.[region].amazonaws.com
15. Put the cloudfront url https://[origin].cloudfront.net into the Custom public URL field.

**Actual behavior:**

When listmonk bucket type is set to private:
- Uploading of media works
- Serving of media does not work. The URLs generated are presigned URLs linking directly to the S3 bucket, but those will not work because the bucket is private. All media URL links are broken.

When listmonk bucket type is set to public:
- Uploading of media does not work. There is a permissions error. 
- Serving of media works. The custom public URL is used correctly to generate working media URLs.

**Expected behavior:**

- Listmonk bucket type setting should be set to private because the bucket is private.
- Uploading of new media should work.
- URLS for media should be generated correctly and should serve correctly using the custom public (CloudFront CDN) URL.

rhnvrm:
Fixes #1141

### DIFF
--- a/internal/media/providers/s3/s3.go
+++ b/internal/media/providers/s3/s3.go
@@ -81,8 +81,9 @@ func (c *Client) Put(name string, cType string, file io.ReadSeeker) (string, err
 
 // Get accepts the filename of the object stored and retrieves from S3.
 func (c *Client) Get(name string) string {
-	// Generate a private S3 pre-signed URL if it's a private bucket.
-	if c.opts.BucketType == "private" {
+	// Generate a private S3 pre-signed URL if it's a private bucket, and there
+	// is no public URL provided.
+	if c.opts.BucketType == "private" && c.opts.PublicURL == "" {
 		u := c.s3.GeneratePresignedURL(simples3.PresignedInput{
 			Bucket:        c.opts.Bucket,
 			ObjectKey:     c.makeBucketPath(name),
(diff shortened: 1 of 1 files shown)


rhnvrm:
With custom URL set, generate URL with Public URL.

image

image

Without custom URL set, generate resigned URL

image

Issue #241, opened 2020-12-14
email attachments?


praul:
Hi and first of all, thanks for listmonk, I really like the clean , decluttered approach.

I need to send emails with pdfs as attachment.
Is that possible, supported or planned?

Thanks for answering

knadh:
Thanks @praul. Attachments aren't immediately planned as they're an unlikely usecase in newsletters and mailing lists. They can also significantly affect bandwidth usage and performance of campaigns.

Trapulo:
I'm also interested in this..

thanks

m42e:
What is the benefit of sending the PDF via mail than to add a Tracking Dowload Link?

Trapulo:
You do not need a storage where leave them available...

Empty2k12:
We're also interested in this feature. When sending out press releases we sometimes like to attach the press release as a PDF or add additional images that should not appear in the e-mail test itself. It's a standard practice to attach things to e-mails in this context.

knadh:
I'll look into this once after the next major release (bounce processing). This is not a trivial feature to implement. Re-opening the issue for now.

avanier:
I'm going to agree, this feature is a bit of a slippery slope, but I can also add other potential use-cases.

Our use is internal and we could really use attaching `.ics` files or meeting invites to certain campaigns.

xam-ps:
I use listmonk to send out newsletter for our fire department and here it would also be quite helpful to be able to attach files.

knadh:
I've looked into this and enabling native attachments poses several technical challenges.

- When an attachment is uploaded to a campaign, it has to be stored somewhere. The existing image/media upload backend can be used for this.
- When a campaign is being processed, the attachments have to be read and held in memory. Reading from disk or S3 per message is not practical. However, large attachments could pose problems when held in-memory.
- How can file deletion be handled (from the media gallery) when there's a running campaign or a paused campaign?

What is easier is to refactor the media upload feature to accept arbitrary file types and not just images. Then the richtext toolbar can then get a button that opens up the file gallery that inserts the link to the desired file in the campaign body. Not an e-mail attachment, but simpler UX for uploading and linking files from within listmonk.

NicoHood:
But keep in mind: This would make the attachment public visible to anyone with the link. E-Mail attachments are just sent 1:1 normally. That'swhy I'd say this is a campaign-specific upload and should not be merged with the media library?

* We need a new backend/api
* The file deletion problem will be solved.

[...]

PR #1166, opened 2023-02-01
Enable TxMessage attachments through API 

lidakaml:
This implementation allows for attachments in transactional messages.  
Tested with Email messenger.  
No additional validation, since file/name should be checked by API client.  

Helpful for sending invoices, tickets etc.

Semi fixes issue #241

### DIFF
--- a/internal/manager/manager.go
+++ b/internal/manager/manager.go
@@ -411,16 +411,7 @@ func (m *Manager) worker() {
 				return
 			}
 
-			err := m.messengers[msg.Messenger].Push(messenger.Message{
-				From:        msg.From,
-				To:          msg.To,
-				Subject:     msg.Subject,
-				ContentType: msg.ContentType,
-				Body:        msg.Body,
-				AltBody:     msg.AltBody,
-				Subscriber:  msg.Subscriber,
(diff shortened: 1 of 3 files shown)


internal/manager/manager.go
```diff
@@ -411,16 +411,7 @@ func (m *Manager) worker() {
...
-				From:        msg.From,
-				To:          msg.To,
-				Subject:     msg.Subject,
-				ContentType: msg.ContentType,
-				Body:        msg.Body,
-				AltBody:     msg.AltBody,
-				Subscriber:  msg.Subscriber,
-				Campaign:    msg.Campaign,
-			})
+			err := m.messengers[msg.Messenger].Push(msg.Message)
```
lidakaml, line 414: This achieves the same, but shorter

knadh:
Thanks for the PR. Was unable to push changes to your branch, so had to do a new commit. https://github.com/knadh/listmonk/commit/55f7eca2e8a81a8b6ae1e79386b2eef18104dc44

Instead of the Base64 encoded files as a part of the JSON payload, it now accepts multipart form posts.

PR #1168, opened 2023-02-02
[i18n] Add translation of the term Powered by

kosssi:
Hello,

I would like add tranlsation for the term "Powered by".

Thanks for listmonk ❤️


kosssi:
thanks @knadh I change the key ;)

Issue #929, opened 2022-08-30
Importing subscribers who were initially blocklisted during a previous import


Kannaj:
Hi Kailash,

**TLDR**: we see a bug when we upload a bunch of users (previously blocklisted) to a new list. Ideally, they should still be updated as 'unsubscribed' in the new list 

-------

We recently moved our newsletter arch from email octopus to listmonk. This process involved generating a 'subscribed' list and an 'unsubscribe' list from email octopus. FYI, these lists are not mutually exclusive and there is some overlap.

Our import procedure involved the below

1. Import the new list of subscribers as `confirmed` in `/admin/subscribers/import`
2. Blocklist a portion of those users (and add additional ones)  

The problem happens when we import a new list of subscribers from our prod db and if those users happen to already be marked as blocklisted in listmonk db.

These users are marked as 'confirmed' in the `subscriber_list` table but are `blocklisted` at the subscriber level. This means that they're essentially counted in the `subIDs` CTE but filtered out in the `subs` CTE

```
WITH camps AS (
    SELECT last_subscriber_id, max_subscriber_id, type FROM campaigns WHERE id = $1 AND status='running'
),
campLists AS (
    SELECT lists.id AS list_id, optin FROM lists
    LEFT JOIN campaign_lists ON (campaign_lists.list_id = lists.id)
    WHERE campaign_lists.campaign_id = $1
),
subIDs AS (
    SELECT DISTINCT ON (subscriber_lists.subscriber_id) subscriber_id, list_id, status FROM subscriber_lists
    WHERE
        -- ARRAY_AGG is 20x faster instead of a simple SELECT because the query planner
        -- understands the CTE's cardinality after the scalar array conversion. Huh.
        list_id = ANY((SELECT ARRAY_AGG(list_id) FROM campLists)::INT[]) AND
        status != 'unsubscribed' AND
        subscriber_id > (SELECT last_subscriber_id FROM camps) AND
        subscriber_id <= (SELECT max_subscriber_id FROM camps)
    ORDER BY subscriber_id LIMIT $2
),
subs AS (
    SELECT subscribers.* FROM subIDs
    LEFT JOIN campLists ON (campLists.list_id = subIDs.list_id)
    INNER JOIN subscribers ON (
        subscribers.status != 'blocklisted' AND
        subscribers.id = subIDs.subscriber_id AND

        (CASE
            -- For optin campaigns, only e-mail 'unconfirmed' subscribers.
[...]

PR #1175, opened 2023-02-04
Fix #1155. Empty output subscribers query, who finish campaign

### LINKED ISSUE #1155: Campaign goes to finished status without reaching all subscribers
**Version:**
 - listmonk: 2.3.0
 - OS: Ubuntu 22.04.1 LTS
 - Docker: Docker version 20.10.22, build 3a2c30b

Faced the problem that the campaign ends after reaching only 3 subscribers out of 786210.

I checked several subscribers - all fit the conditions of sending.

Campaign data from database:
```
select id, uid, send_at, status, type, messenger, to_send, sent, max_subscriber_id , last_subscriber_id , started_at , created_at , updated_at
from campaigns c
where id = 23

Name              |Value                               |
------------------+------------------------------------+
id                |23                                  |
uuid              |5cd4919a-9fe9-4adc-a567-452b3dc92402|
send_at           |                                    |
status            |finished                            |
type              |regular                             |
messenger         |email                               |
to_send           |786210                              |
sent              |3                                   |
max_subscriber_id |5311356                             |
last_subscriber_id|28                                  |
started_at        |2023-01-25 13:29:40.477 +0300       |
created_at        |2023-01-25 12:51:01.551 +0300       |
updated_at        |2023-01-25 13:29:42.611 +0300       |
```

Log:
```
listmonk_app  | 2023/01/25 10:29:42 manager.go:510: start processing campaign (Name)
listmonk_app  | 2023/01/25 10:29:42 manager.go:684: campaign (Name) finished
```

Can you please tell me how to find the cause of the problem?

saginsa:
Fix problem Campaign goes to finished status without reaching all subscribers #1155. 

It was in us, when we import big blacklisted base at start work with listmonk, and future lists have blocklisted subscribers but they wasn't unsubscribed. It was problem, because at first sample in query, find only unsubscribed emails, but in finish all unsubscribed finish transform in empty, because all of them is blacklisted. This element stop campaign, despite the remaining large number of emails on the mailing list, despite the remaining large number of emails on the mailing list.

### DIFF
--- a/queries.sql
+++ b/queries.sql
@@ -689,30 +689,27 @@ campLists AS (
     WHERE campaign_lists.campaign_id = $1
 ),
 subIDs AS (
-    SELECT DISTINCT ON (subscriber_lists.subscriber_id) subscriber_id, list_id, status FROM subscriber_lists
+    SELECT DISTINCT ON (sl.subscriber_id) sl.subscriber_id, sl.list_id, sl.status FROM subscriber_lists sl
+    INNER JOIN subscribers ON (subscribers.id = sl.subscriber_id AND subscribers.status!='blocklisted')
     WHERE
         -- ARRAY_AGG is 20x faster instead of a simple SELECT because the query planner
         -- understands the CTE's cardinality after the scalar array conversion. Huh.
-        list_id = ANY((SELECT ARRAY_AGG(list_id) FROM campLists)::INT[]) AND
-        status != 'unsubscribed' AND
-        subscriber_id > (SELECT last_subscriber_id FROM camps) AND
-        subscriber_id <= (SELECT max_subscriber_id FROM camps)
-    ORDER BY subscriber_id LIMIT $2
+        sl.list_id = ANY((SELECT ARRAY_AGG(list_id) FROM campLists)::INT[]) AND
+        sl.status != 'unsubscribed' AND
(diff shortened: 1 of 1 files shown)


saginsa:
I think, that this fix haven't big performanse degradate, because subIds select work in same logical framework and conditions, and this is logical true.
Also it fix #929 , without create multiple unsubscribed records on all lists in listmonk for all blocklisted user.

knadh:
Thank you. Will test this on a large  DB with millions of records over the weekend.

queries.sql
```diff
@@ -689,30 +689,27 @@ campLists AS (
     WHERE campaign_lists.campaign_id = $1
 ),
 subIDs AS (
-    SELECT DISTINCT ON (subscriber_lists.subscriber_id) subscriber_id, list_id, status FROM subscriber_lists
+    SELECT DISTINCT ON (sl.subscriber_id) sl.subscriber_id, sl.list_id, sl.status FROM subscriber_lists sl
+    INNER JOIN subscribers ON (subscribers.id = sl.subscriber_id AND subscribers.status!='blocklisted')
```
knadh, line 693: Instead of doing a join on subscribers here and another later, it should ideally be just one.

saginsa: I remade it. Please take a look. 

Also i removed filter campLists.optin = 'double' in opt-in campaign. 
I think that this is logical mistake.

PR #1217, opened 2023-02-26
Svg images support

eltorio:
Now svg images are supported in nearly all email clients and have many advantages for logo for example.
This adds support for svg image in uploads


knadh:
Thanks for the PR. I checked Gmail and it doesn't support inserting or rendering SVGs. Thunderbird does. Could you share some popular e-mail clients that support SVG?

eltorio:
Hi,
At least
Proton Mail does,
Apple Mail 16 too,
https://outlook.office.com/ also,

cmd/media.go
```diff
@@ -150,11 +162,25 @@ func handleDeleteMedia(c echo.Context) error {
...
 
 	app.media.Delete(fname)
-	app.media.Delete(thumbPrefix + fname)
+	app.media.Delete(thumbPrefix + fname + addPrefixIfNeeded(fname))
 
 	return c.JSON(http.StatusOK, okResp{true})
 }
 
+// generate a png thumbnail of src
+func getSVGThumb(src multipart.File) (image.Image, error) {
```
knadh, line 171: I think we don't need this at all. SVG is vector, so for thumbnails, it can just be scaled down and rendered. The thumb path and the full path can be the same image.

PR #1262, opened 2023-03-25
Public Archive Improvements

### LINKED ISSUE #1033: Campaign body in RSS feed
Currently the RSS feed only contains the title and people have to click through to read the post.

Scherm­afbeelding 2022-11-16 om 20 04 20

It would be great to have the whole post in the RSS feed.

justinbeaty:
This PR does two things. I can split into separate PRs if requested.

1. Adds an `example.com/archive/latest` page that renders the most recent newsletter.
2. Adds a `<content:encoded>` element into the RSS feed. (Fixes #1033)

For item 2, perhaps we should add a config option for this? Something like this (mockup):

image

Note: I did have to add complexity to the `get-archived-campaigns` query, which I think is an acceptable trade off and it makes the function's description more accurate:

https://github.com/knadh/listmonk/blob/684c15a4049517abcad4768122498163b0f761bd/internal/core/campaigns.go#L140-L141


knadh:
This is nice! Thanks @justinbeaty. Related: https://github.com/knadh/listmonk/issues/1033

Yes, we should have that as a config option. Would you be able to add a commit for that or would you want me to? We've to:

- Add a DB migration for v2.5.0 for the new config field.
- Add it to the Vue UI and in settings.go
- Add it to `constants` (init.go) for it to be available to the handler to decide.

justinbeaty:
> We've to:
> 
> * Add a DB migration for v2.5.0 for the new config field.
> * Add it to the Vue UI and in settings.go
> * Add it to `constants` (init.go) for it to be available to the handler to decide.

Thanks for the list of TODO, I can add the commit.

justinbeaty:
Added the config option with default value `true`. I guess that's probably okay, but could change it to `false` if you think that's better.

justinbeaty:
I've pushed this code to production for a few days now and all seems well. I've been testing the public rss feed with inoreader.com.

knadh:
Thanks @justinbeaty. Will review this and merge this in the coming few days.

knadh:
Hi @justinbeaty. I've pushed a couple of commit with minor changes. Can you review?

justinbeaty:
Fixed a typo in `i18n/en.json` ("Show ___full full___ e-mail content") and re-ran `scripts/refresh-i18n.sh`.

I will test this out today. For my own reference, I ran these postgres queries so I can re-run the update process:

```sql
update settings set value = '["v1.0.0", "v2.0.0", "v2.1.0", "v2.2.0", "v2.3.0", "v2.4.0"]' where key = 'migrations';
delete from settings where key = 'app.enable_rss_content';
```

internal/migrations/v2.5.0.go
```diff
@@ -11,7 +11,7 @@ func V2_5_0(db *sqlx.DB, fs stuffbin.FileSystem, ko *koanf.Koanf) error {
 	// Insert new preference settings.
 	if _, err := db.Exec(`
 		INSERT INTO settings (key, value) VALUES
- 			('app.enable_rss_content', 'true')
+ 			('app.enable_public_archive_rss_content', 'false')
```
justinbeaty, line 14: You've changed this to `false`, but in `schema.sql` we have:
```
('app.enable_public_archive_rss_content', 'true')
```

Maybe it's intentional to give different defaults when installing fresh vs upgrading?

justinbeaty:
@knadh Besides the one comment above about default true/false values, all seems good. I tested locally and everything was working.

internal/migrations/v2.5.0.go
knadh: Ah yes, forgot to document this. The existing installations that use the archive feature would expect the content not to be published (as there's no support now). We don't want to break this expectation+behaviour by publishing full content after an upgrade. This should be an explicit choice for existing installation.

justinbeaty: @knadh agreed, that’s reasonable decision.

knadh:
Thank you once again for the excellent PR @justinbeaty!

Sjors:
Thanks for this! Are you planning on tagging a new release in the near future?

justinbeaty:
@Sjors per the comment [here](https://github.com/knadh/listmonk/pull/1385#issuecomment-1623013121) it would seem there’s a release coming. Hope you enjoy it.

Issue #1275, opened 2023-04-06
Subdomain wildcards in blocklist


boenner:
In some cases it would be helpful to blocklist all subdomains with a wildcard to avoid sending any mails to a certain organization. For example, a lot of universities use subdomains for faculties in their mail addresses.

PR #1276, opened 2023-04-06
Subdomain wildcards in blocklist

boenner:
This relates to #1275. Enables the domain blocklist to use a `*` wildcard prefix to match all subdomains for a domain.


knadh:
Thanks for the PR. I merged a different approach that supports N levels of subdomains and avoid map iteration on every check. Eg: *.a.b.example.com

PR #1295, opened 2023-04-18
add translation status to README

felixhaeberle:
Hey there, we want to get you more contributors to your repository!
Therefore, I added a badge which lets users contribute to your translations.

The badge endpoint allows you to create a dynamic badge that display real-time data on your project. For instance, you can use it to showcase your project's localization progress or the number of errors/warnings on your Github repository. This means that whenever someone views the badge, they will see the most up-to-date information about your project.

For your project, the badge looks like this:
(this image is already auto generated)

[translation badge](https://inlang.com/editor/github.com/knadh/listmonk?ref=badge)

We would be really happy if you decide to place this into your README to let people contribute to your translations 🥰

– Thank you,
Felix

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,10 @@ __________________
 
 
 ## Developers
-listmonk is a free and open source software licensed under AGPLv3. If you are interested in contributing, refer to the [developer setup](https://listmonk.app/docs/developer-setup). The backend is written in Go and the frontend is Vue with Buefy for UI. 
+listmonk is a free and open source software licensed under AGPLv3. If you are interested in contributing, refer to the [developer setup](https://listmonk.app/docs/developer-setup). The backend is written in Go and the frontend is Vue with Buefy for UI.
+
+### Translation status
+[![translation badge](https://inlang.com/badge?url=github.com/knadh/listmonk)](https://inlang.com/editor/github.com/knadh/listmonk?ref=badge)
 
 
 ## License


knadh:
Because all language files in listmonk will have all the keys as the base language (English) including untranslated ones with default language (English), this badge will always show 100%, right?

Eg: https://github.com/knadh/listmonk/blob/master/i18n/zh-CN.json#L388

felixhaeberle:
Hey 👋 No, we specifically lint all the property's and look for values / lint rules applied. The translation status percentage is based on the existence of a string and if all the specified lint rules are satisfied. 

For example you can force a specific format of a translated string with a lint rule (basically a function which checks existence or format). If this format isn't correct, this badge will display a warning / error and the percentage will not be 100% because now 1 reference is missing (existence or format).

The badge is a nice analytics feature where you can always see if everything is correct and guide contributor to your translations if something is translated wrong / a translation needs to be added.

knadh:
Hi @felixhaeberle. The badge still shows 100% currently when many language files still have untranslated, default English strings.

felixhaeberle:
@knadh Oh now I get it, I just saw that you are using english strings in the chinese translation. Unfortunately, that's not how it's designed to work on our side. But you can benefit from the errors and warnings in the future — let me know what you think! :)

felixhaeberle:
@knadh Hey, just a small check-in on this 😇 
Have you already decided if you want to have this badge in the case of potential lint errors / warnings?

felixhaeberle:
@knadh We have created a lint rule from which you can benefit and find dupicates in different languages – we display them as warnings, which you can also supress.

https://github.com/inlang/standard-lint-rules/issues/6

knadh:
The badge still shows 100%. I don't think a global number for all languages put together is useful in listmonk's context. Maybe a badge that shows a table of every language package and the respective statuss would be useful to potential translators. Thank you.

PR #1309, opened 2023-05-07
Enable extra system calls in systemd service

runningnoodle:
Proposed fix for #1310. 

Since v2.4.0, it looks like Listmonk is making system calls for changing resource limits, memory or scheduling parameters. This PR allows those system calls in the systemd service configuration. This allows Listmonk to run as a systemd service.

The downside of this change is that it weakens the sandbox around the Listmonk process when run as a systemd service. I don't fully understand how systemd works, but I don't expect any serious consequence.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,9 +1,7 @@
 name: goreleaser
 
-on:
-  push:
-    tags:
-      - "v*" # Will trigger only if tag is pushed matching pattern `v*` (Eg: `v0.1.0`)
+on: workflow_dispatch
+
 
 jobs:
   goreleaser:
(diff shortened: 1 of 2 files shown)


knadh:
@runningnoodle this is curious. There have been no changes in v2.4.0 that should do this. No library that should also exhibit this behaviour. Would like to figure out this behaviour somehow before we merge this.

runningnoodle:
I closed by accident. The last two commits were also accidents. Sorry about the mess I'm creating.

ozgurerdogan:
When trying to start service, I am getting;
init.go:798: error starting HTTP server: listen tcp 0.0.0.0:392: bind: permission denied

PR #1327, opened 2023-05-13
Documentation: how to enable uploads when not using docker

dc:
I run the listmonk binary directly without docker. Uploads always failed and the error messages disappeared so quickly that I could not read it. I found out by trial and error that simply creating a `uploads` folder is enough to fix this.

So here’s my documentation extension that explains how to enable uploads when not using docker.


knadh:
Hi @dc. You're supposed to configure this in the settings. It doesn't need to be `uploads`, can be any directory of your choice.

image

PR #1341, opened 2023-05-17
Add support for file attachments on campaigns

knadh:


avanier:
Just linking the issue. This addresses #241 .

PR #1344, opened 2023-05-23
feat: add multiarch docker build support

Thunderbottom:
- Added docker build support for the following platforms:
  - amd64 (default)
  - arm64v8
  - armv6
  - armv7
- Added `GOOS` and `GOARCH` information to the `buildString`
- Added docker manifest details to store image architecture metadata
- Updated `docker/login-action` from `v1` to `v2`
- Updated `goreleaser/goreleaser-action` from `v2` to `v4`

Tested the images locally, built with `goreleaser` and seems to be working fine. Screenshots:

#### Build Images
image

#### Running Containers
image

#### Container Binary Build Info
image

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -24,7 +24,7 @@ builds:
       - 6
       - 7
     ldflags:
-      - -s -w -X "main.buildString={{ .Tag }} ({{ .ShortCommit }} {{ .Date }})" -X "main.versionString={{ .Tag }}"
+      - -s -w -X "main.buildString={{ .Tag }} {{ .Release.Goos }}/{{ .Release.Goarch }} ({{ .ShortCommit }} {{ .Date }})" -X "main.versionString={{ .Tag }}"
 
     hooks:
       # stuff executables with static assets.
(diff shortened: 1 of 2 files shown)


.goreleaser.yml
```diff
@@ -37,15 +37,16 @@ archives:
...
 
 dockers:
-  - goos: linux
+  - use: buildx
+    goos: linux
     goarch: amd64
     ids:
     - listmonk
     image_templates:
-    - "listmonk/listmonk:latest"
```
knadh, line 45: Does this mean that a `docker pull listmonk/listmonk:latest` will no longer work or will the `-amd64` bit resolve automatically?

Thunderbottom: docker automatically resolves this based on the system's architecture. so running `docker pull listmonk/listmonk:latest` on an `amd64` machine will pull the `amd64` image, whereas on `arm` it'll automatically pull the `arm` version of the image.

Thunderbottom: source: https://docs.docker.com/build/building/multi-platform/#building-multi-platform-images

activenode:
This is desperately needed :) 

Thanks for the PR. Wanted to do the exact same thing <3 

Can we have this merged soon?

knadh:
Done. Thanks @Thunderbottom!

activenode:
Neat. One quick question. I am not seeing auto-releases for main pushes. Only for `v*` tags. So do you have in mind when to push the next tag so that it gets pushed to DockerHub? :)

Thanks a bunch @knadh for this great piece.

knadh:
Hi @activenode. I don't think it's ideal for the entire package to be built and pushed on every commit, where some commits may be trivial. A better approach would be to have an automated monthly RC release or something along those lines.

activenode:
Sure that's fine. I just wanted to ask basically when the next version release is expected. Until then I can build it on my own, all good

dandry:
Adding to what @activenode asked. @knadh is there any schedule for the release? I am so eager to try listmonk on my ARM VPS. Great project by the way. Thank you for contributing to it.

activenode:
@dandry I created an arm64 build because I wouldn't want to wait. If you wanna try it. It's the exact latest version. https://hub.docker.com/repository/docker/activenode/listmonk-arm64/tags?page=1&ordering=last_updated

dandry:
@activenode thank you, will give it a try!

activenode:
Coming back to this thread, I was forking the newest main and seeing that `goreleaser` wouldn't do all the builds now as it fails with `darwin` and `windows` . Which is totally okay because who wants that anyway :D buuuut it breaks the build really. 

I'm working on that in my fork and will provide a PR back.

activenode:
Little Update:

I'm still working on my branch, multiple errors occur of which one is that syscall.Utsname isn't available on all platforms. I am trying to find a more fitting/generic solution

activenode:
https://github.com/knadh/listmonk/pull/1414

PR #1373, opened 2023-06-13
Improved CONTRIBUTING.md

Harshad112:
Made changes so user can see neat and clean


knadh:
Thanks for the PR, but this is a bit of a superfluous change which I don't think is necessary.

PR #1453, opened 2023-08-13
Better Feedback when trying SMTP Test with existing creds

activenode:
Since the SMTP Test is only working for unsaved passwords (as otherwise the password is just ******) I would at least love to give that feedback to the user.

If you feel this is the wrong approach and we rather should then read it from the database and run the Test Mail, fine by me as well HOWEVER I think we need a better UX either way to make it clear what's happening.

So if you think this solution is good in the meantime providing proper Error message, please merge :)

### DIFF
--- a/cmd/settings.go
+++ b/cmd/settings.go
@@ -262,6 +262,17 @@ func handleTestSMTPSettings(c echo.Context) error {
 		return echo.NewHTTPError(http.StatusBadRequest, app.i18n.Ts("globals.messages.missingFields", "name", "email"))
 	}
 
+	password := ko.String("password")
+	// now we have to check if the password only contains •••••••••••••••• chars because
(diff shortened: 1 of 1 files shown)


knadh:
https://github.com/knadh/listmonk/blob/f3e9b08026d6639e72382f7b2efa7b9c3fecc6cf/frontend/src/views/settings/smtp.vue#L301

The frontend shows a warning. I just realised that in the new release, [this condition broke](https://github.com/knadh/listmonk/blob/f3e9b08026d6639e72382f7b2efa7b9c3fecc6cf/frontend/src/views/settings/smtp.vue#L334) ;( Instead of checking for empty spaces, it should be checking for the dummy password. Will fix and push another release soon as this is going to be very annoying for users.

activenode:
Ok then feel free to close this

Issue #1473, opened 2023-08-21
Modify Install Script to support MacOS Ventura


hdmoreland:
**Issue**

The current install instructions do not work correctly on MacOS Ventura because the sed commands in the ```install-prod.sh``` script is for GNU sed, but MacOS uses BSD sed. 

This was reported previously in #1083, where the user manually switched their system to use gnu-sed. Other OS-compatibility issues have been reported in the past (#624) and have subsequently been fixed (#625).

**Steps to Replicate**

Run the current install script on MacOS Ventura:

```
(base) holden@mbp listmonk % mkdir listmonk && cd listmonk
sh -c "$(curl -fsSL https://raw.githubusercontent.com/knadh/listmonk/master/install-prod.sh)"

>  checking for an existing docker db volume
>  fetching config.toml from listmonk repo
>  fetching docker-compose.yml from listmonk repo
>  generating a random password
>  modifying config.toml
sed: 1: "config.toml": command c expects \ followed by text
(base) holden@@mbp listmonk %
```

**Solution**

Modify the install-prod.sh script to check if the Operating system is MacOS on install. Modify the sed command to function based on the user's OS. 

A new function can be created to determine the proper sed command:
```
configure_sed(){
	# Detect the platform
	if [[ "$(uname)" == "Darwin" ]]; then
		# macOS sed
		SED_INPLACE="-i ''"
	else
		# Assume GNU sed
		SED_INPLACE="-i"
	fi
}
```

The ```modify_config``` function can be altered to use this variable:
```
modify_config(){
	info "generating a random password"
	db_password=$(generate_password)

	info "modifying config.toml"
	# Replace `db.host=localhost` with `db.host=db` in config file.
	sed $SED_INPLACE "s/host = \"localhost\"/host = \"listmonk_db\"/g" config.toml
	# Replace `db.password=listmonk` with `db.password={{db_password}}` in config file.
	# Note that `password` is wrapped with `\b`. This ensures that `admin_password` doesn't match this pattern instead.
	sed $SED_INPLACE "s/\bpassword\b = \"listmonk\"/password = \"$db_password\"/g" config.toml
	# Replace `app.address=localhost:9000` with `app.address=0.0.0.0:9000` in config file.
	sed $SED_INPLACE "s/address = \"localhost:9000\"/address = \"0.0.0.0:9000\"/g" config.toml

	info "modifying docker-compose.yml"
[...]

PR #1474, opened 2023-08-21
Modify sed command based on OS

hdmoreland:
Fixes issue #1473 by modifying the sed command used in the install script based on the OS.

### DIFF
--- a/install-prod.sh
+++ b/install-prod.sh
@@ -44,6 +44,17 @@ check_dependencies() {
 	fi
 }
 
+configure_sed(){
+	# Detect the platform
+	if [[ "$(uname)" == "Darwin" ]]; then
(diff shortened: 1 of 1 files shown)


tosie:
+1. Looks solid from my side and can confirm the differences in `sed` ...

install-prod.sh
```diff
@@ -44,6 +44,17 @@ check_dependencies() {
 	fi
 }
 
+configure_sed(){
```
knadh, line 47: I think we can simplify this with a one line definition at the beginning of the script along with other globals so that it's explicit.

`SED_INPLACE=$(if [[ "$(uname)" == "Darwin" ]]; then echo "-i ''"; else echo "-i"; fi)`

PR #1476, opened 2023-08-22
Making Dockerfile more flexible for builds in different CPU arch

tigranbs:
Using Listmonk on Arm64 from Docker is quite complex unless you build a binary and copy it to the container, which is a lot more work than just running the `docker build` command.
This PR makes Dockerfile more flexible to just run Docker Build with all of the build processes that Listmonk has, which comes with a benefit that if you need to build for Arm64, then use the Docker cross arch builder command.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,28 @@
+FROM node:19-alpine as front-builder
+
+RUN apk add --no-cache alpine-sdk
+
+WORKDIR /app
+ADD . /app
+
+RUN make build-frontend
+
+FROM golang:alpine as builder
+
+RUN apk add --no-cache alpine-sdk
+
+WORKDIR /app
+ADD . /app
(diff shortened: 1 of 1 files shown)


knadh:
I think this breaks the goreleaser build as it expects to execute the build steps (outside of the Dockerfile).

knadh:
Could you please confirm the goreleaser behaviour?

tigranbs:
I tried it on my Macbook M1 Max, and with this PR, `docker build` works as expected, and it produces cross-platform builds for Arm64 and AMD64 Linux machines.

```
docker buildx build --platform linux/amd64 -t repo.alin.io/tigranbs/listmonk:latest --push .
```

knadh:
That would work, but the GoReleaser functionality may not. Please try `make release-dry`.

knadh:
Closing this as it's incompatible with goreleaser. Thanks.

PR #1489, opened 2023-08-28
fix: Replace deprecated docker-compose commands with docker compose

### LINKED ISSUE #1431: End of support for Compose V1
Using the `install-prod.sh` script on a clean installed server with latest docker fails because it checks for docker-compose command which doesn't not exist for my version (`Docker version 24.0.2, build cb74dfc`).

replace all `docker-compose` by docker compose.

Should i open a PR for that ? `install-demo.sh` would also need to be updated

Madscientiste:
Corrected the setup & demo scripts by replacing outdated `docker-compose` commands with the appropriate docker compose commands for Listmonk setup, also eliminated checks for `docker-compose` command.

I have tested both scripts inside a `dev-container`, if something isn't right, let me know 👍 

Should close #1431

### DIFF
--- a/install-demo.sh
+++ b/install-demo.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env sh
 set -eu
 
-# Listmonk demo setup using `docker-compose`.
+# Listmonk demo setup using `docker compose`.
 # See https://listmonk.app/docs/installation/ for detailed installation steps.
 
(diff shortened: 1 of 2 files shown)


knadh:
This was done in https://github.com/knadh/listmonk/pull/1490. Thank you!

PR #1507, opened 2023-09-04
refractor: handleGetLists function

amansaini0003:
Refractor handleGetLists function by removing variable.


knadh:
hm, thanks, but this change makes no sense. The `var()` approach for a group of variables in one block is a pattern the project follows.

amansaini0003:
> hm, thanks, but this change makes no sense. The `var()` approach for a group of variables in one block is a pattern the project follows.

okay I can revert that, but there is no need for a `single` variable in `handleGetLists` that was redundant.

Issue #1509, opened 2023-09-05
Query, ANY keyword doesn't work


FournyP:
**Version:**
 - listmonk: 2.4.0
 - OS: Windows 11 (WSL 2)

**Description of the bug and steps to reproduce:**
I try to get some lists using the API thanks to the `query` parameter.
My query is very simple : `'alerts' = ANY(lists.tags)` in order to find lists that contains the tag `alerts`

Firstly, the request didn't work and I found this issue : #1471.
So I've added `\` before parentheses and it worked, I got a 200 status. 
But the result is not what I expected, it's not the same result as the database (see screenshots)

**Screenshots:**
![image](https://github.com/knadh/listmonk/assets/64586968/1ad539f2-5f36-4797-a6bc-ed812a4dae7b)
![image](https://github.com/knadh/listmonk/assets/64586968/bf799cba-60a1-4ebe-9b2b-734fe0ab3d68)

As you can see, the result is empty but in my database, I got one row with the `alerts` tag

Here the request that I use : `http://localhost:9000/api/lists?query=%27alerts%27%3DANY%5C%28lists.tags%5C%29`

If you want to reproduce, here my listmonk container from my docker-compose :
```
listmonk:
    image: listmonk/listmonk:v2.4.0
    container_name: ezexchange.api.listmonk
    ports:
      - "9000:9000"
    networks:
      - listmonk-tier
      - database-tier
    environment:
      - TZ=Etc/UTC
    command: # Comment this config if you don't want to initialize the database
      [
        sh,
        -c,
        "yes | ./listmonk --install --config /listmonk/config.toml && ./listmonk --config /listmonk/config.toml",
      ]
    depends_on:
      db:
        condition: service_healthy
    volumes:
      - ./listmonk.toml:/listmonk/config.toml
```

And I've just add the `alerts` tag manually in the database

Have a good day ;)

FournyP:
I just try using listmonk 2.5.1 and it also don't work.

I try also this expression : `lists.tags @> '{alerts}'` using this request `http://localhost:9000/api/lists?query=lists.tags%20%3D%20%27%7Balerts%7D%27` but I got a error message :
![image](https://github.com/knadh/listmonk/assets/64586968/03f04bc6-5109-4060-8da7-69b2920c34c9)

I have the felling that the query don't take in account `{` and `}` according to the response

knadh:
Hi @FournyP. Apologies for the delay in responding.

[...]

PR #1510, opened 2023-09-05
Update QueryLists to request using custom query

FournyP:
After my issue #1509, I realised that the `query` parameter in order to query lists is just used for the `name` column from the list database...

So we can't query on `tags` and on other attributes. 

I found that weird because the `query` param for subscribers had this functionnality.

So I've made this little improvement, ~~copy~~ strongly inspired from the subscriber feature.

Here the result before the improvement :
image

Here the result after the improvement :
image

Hope this issue will be merge because I really need it
Feel free to ask me modifications

Have a good day


knadh:
Closing this. Please see https://github.com/knadh/listmonk/issues/1509#issuecomment-1724873144.

PR #1529, opened 2023-09-19
fix: update inlang settings

jannesblobel:
Fixes # (issue)
Currently inlang does not work for Cal.com. This PR fixes this problem, and the VS extension and editor will again work as expected.

Test yourself on:
https://inlang.com/editor/github.com/jannesblobel/listmonk


knadh:
Hi @jannesblobel. Would it be possible to introduce an `.inlang` directory to which the files can be moved? Two meta files for the same tooling in the root directory of a project is a bit noisy.

jannesblobel:
Hey @knadh,I have replaced the `inlang.config.js` with `project.inlang.json`. But you mentioned a good point. Inlang should introduce a `.inlang` directory

PR #1554, opened 2023-10-11
updated spanish and italian translations

jorge-vitrubio:


knadh:
Thanks for the PR, but I don't think it's necessary to specify `SiteKey` in one specific language file. It's understood that it's the hCaptcha key + secret that's required.

PR #1605, opened 2023-11-25
Modern Greek translation

waseigo:
Generated on https://listmonk.app/i18n/, with manual tweaking of translations using ChatGPT 3.5 and DeepL.com.


knadh:
Thanks @waseigo! :tada:

Issue #843, opened 2022-06-19
how to install binary as a service in ubuntu - just getting errors


bamboowonder:
Hi there,

I don't have or want docker, so I'm attempting to use the binary as a service on ubuntu 20.04
I found this:  https://github.com/knadh/listmonk/blob/master/listmonk@.service
and tried to adapt it to setup the binary as a service but it continues to fail.

I moved the config to /etc/listmonk and the binary to /usr/bin

after enabled, daemon-reload, start are run, I get this in the log:
`
sudo systemctl status listmonk.service 
● listmonk.service - listmonk mailing list and newsletter manager ()
     Loaded: loaded (/etc/systemd/system/listmonk.service; enabled; vendor preset: enabled)
     Active: failed (Result: exit-code) since Sun 2022-06-19 15:06:53 PDT; 4min 44s ago
  Condition: start condition failed at Sun 2022-06-19 15:11:35 PDT; 2s ago
             └─ ConditionPathExists=/etc/listmonk/.toml was not met

Jun 19 15:06:53 send systemd[1]: listmonk.service: Control process exited, code=exited, status=203/EXEC
Jun 19 15:06:53 send systemd[1]: listmonk.service: Failed with result 'exit-code'.
Jun 19 15:06:53 send systemd[1]: Failed to start listmonk mailing list and newsletter manager ().
Jun 19 15:06:53 send systemd[1]: listmonk.service: Scheduled restart job, restart counter is at 5.
Jun 19 15:06:53 send systemd[1]: Stopped listmonk mailing list and newsletter manager ().
Jun 19 15:06:53 send systemd[1]: listmonk.service: Start request repeated too quickly.
Jun 19 15:06:53 send systemd[1]: listmonk.service: Failed with result 'exit-code'.
Jun 19 15:06:53 send systemd[1]: Failed to start listmonk mailing list and newsletter manager ().
Jun 19 15:11:32 send systemd[1]: /etc/systemd/system/listmonk.service:48: Invalid device node path 'False', ignoring.
Jun 19 15:11:35 send systemd[1]: Condition check resulted in listmonk mailing list and newsletter manager () being skipped.
`

It seems like it is expecting variables to be passed in  
`ConditionPathExists=/etc/listmonk/.toml `
which should be config.toml but in the file is %i.toml

Any insight about the correct way to install as a service would be appreciated.

Thank you,

bamboowonder:
[...]

PR #1614, opened 2023-12-03
Update listmonk@.service

MaximilianKohler:
https://github.com/knadh/listmonk/issues/843#issuecomment-1836783468

### DIFF
--- a/listmonk@.service
+++ b/listmonk@.service
@@ -24,7 +24,7 @@ RestartSec=5
 # To enable a static dir, add the following 
 # --static-dir /etc/listmonk/static
 # to the end of the ExecStart line above after creating the dir and fetching the files with:
-# wget -O - https://github.com/knadh/listmonk/archive/master.tar.gz | tar xz --strip=2 "listmonk-master/static"
+# mkdir -p /etc/listmonk/static ; wget -O - https://github.com/knadh/listmonk/archive/master.tar.gz | tar xz -C /etc/listmonk/static --strip=2 "listmonk-master/static"
 
 # Create dynamic users for listmonk service instances
 # but create a state directory for uploads in /var/lib/private/%i.
--- a/listmonk@.service
+++ b/listmonk@.service
@@ -1,6 +1,9 @@
+# This service template is designed to have the ability to run more than one instance of listmonk with different configurations but the same service unit.
+# It may not work with older OS's with older systemd versions.
+
 [Unit]
 Description=listmonk mailing list and newsletter manager (%I)
-ConditionPathExists=/etc/listmonk/config.toml
+ConditionPathExists=/etc/listmonk/%i.toml
 Wants=network.target
 # The PostgreSQL database may not be on the same host but if it
 # is listmonk should wait for it to start up.
@@ -11,12 +14,16 @@ Type=simple
 PermissionsStartOnly=true
 EnvironmentFile=-/etc/default/listmonk
 EnvironmentFile=-/etc/default/listmonk-%i
-ExecStartPre=/usr/bin/mkdir -p "/listmonk/uploads"
-ExecStartPre=/usr/bin/listmonk --config /etc/listmonk/config.toml --upgrade --yes
-ExecStart=/usr/bin/listmonk --config /etc/listmonk/config.toml $SYSTEMD_LISTMONK_ARGS
+ExecStartPre=/usr/bin/mkdir -p "/etc/listmonk/uploads"
+ExecStartPre=/usr/bin/listmonk --config /etc/listmonk/%i.toml --upgrade --yes
+ExecStart=/usr/bin/listmonk --config /etc/listmonk/%i.toml $SYSTEMD_LISTMONK_ARGS
 TimeoutStopSec=10
 Restart=on-failure
 RestartSec=5
+
+# To enable a static dir, uncomment the ExecStart line below after creating the dir and fetching the files with:
+# wget -O - https://github.com/knadh/listmonk/archive/master.tar.gz | tar xz --strip=2 "listmonk-master/static"
+# ExecStart=/usr/bin/listmonk --static-dir /etc/listmonk/static
 
 # Create dynamic users for listmonk service instances
 # but create a state directory for uploads in /var/lib/private/%i.
--- a/listmonk@.service
+++ b/listmonk@.service
@@ -26,6 +26,9 @@ RestartSec=5
 # to the end of the ExecStart line above after creating the dir and fetching the files with:
 # mkdir -p /etc/listmonk/static ; wget -O - https://github.com/knadh/listmonk/archive/master.tar.gz | tar xz -C /etc/listmonk/static --strip=2 "listmonk-master/static"
 
+# To enable a log file that persists after restarts, replace the ExecStart= line with:
+# ExecStart=/bin/bash -ce "exec /usr/bin/listmonk --config /etc/listmonk/config.toml --static-dir /etc/listmonk/static >>/etc/listmonk/listmonk.log 2>&1"
+
 # Create dynamic users for listmonk service instances
 # but create a state directory for uploads in /var/lib/private/%i.
 DynamicUser=True
--- a/listmonk@.service
+++ b/listmonk@.service
@@ -21,9 +21,10 @@ TimeoutStopSec=10
 Restart=on-failure
 RestartSec=5
 
-# To enable a static dir, uncomment the ExecStart line below after creating the dir and fetching the files with:
+# To enable a static dir, add the following 
+# --static-dir /etc/listmonk/static
+# to the end of the ExecStart line above after creating the dir and fetching the files with:
 # wget -O - https://github.com/knadh/listmonk/archive/master.tar.gz | tar xz --strip=2 "listmonk-master/static"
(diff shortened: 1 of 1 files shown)


MaximilianKohler:
> Your commit breaks the [service template unit](https://www.freedesktop.org/software/systemd/man/latest/systemd.unit.html). That `%i` is there for a reason. Please try to understand what you modify.

Are you talking about `ConditionPathExists=/etc/listmonk/%i.toml` ? Per the referenced issue https://github.com/knadh/listmonk/issues/843 it causes an error, so [this guide](https://mumaritc.hashnode.dev/how-to-install-listmonk-using-binary-on-ubuntu-2204), discussed in that issue, recommends changing it. 

I didn't try without changing those lines since that person had an issue + the guide says to change them, but it worked according to the guide, so that's why I suggested changing it.

MaximilianKohler:
Good to know. I tagged you in the issue because I assumed you know more about this than me. So please feel free to make specific suggestions/changes.

MaximilianKohler:
Apparently my listmonk service stopped on its own:
```
systemctl status listmonk
● listmonk.service
   Loaded: loaded (/etc/systemd/system/listmonk.service; enabled; vendor preset: disabled)
   Active: failed (Result: start-limit) since Sat 2023-12-09 21:18:45 PST; 23h ago

Dec 09 21:31:37 domain systemd[1]: [/etc/systemd/system/listmonk.service:24] Not an absolute path, ignoring: %S/listmonk-
Dec 09 21:31:37 domain systemd[1]: [/etc/systemd/system/listmonk.service:47] Unknown lvalue 'ProtectControlGroups' in s...ervice'
Dec 09 21:31:37 domain systemd[1]: [/etc/systemd/system/listmonk.service:48] Unknown lvalue 'ProtectKernelTunables' in ...ervice'
Dec 09 21:31:37 domain systemd[1]: [/etc/systemd/system/listmonk.service:56] Invalid device node path 'False'. Ignoring.
Dec 09 21:31:37 domain systemd[1]: [/etc/systemd/system/listmonk.service:58] Unknown lvalue 'RestrictNamespaces' in sec...ervice'
Dec 09 21:31:37 domain systemd[1]: [/etc/systemd/system/listmonk.service:60] Unknown lvalue 'RestrictRealtime' in secti...ervice'
Dec 09 21:31:37 domain systemd[1]: [/etc/systemd/system/listmonk.service:65] Unknown lvalue 'MemoryDenyWriteExecute' in...ervice'
Dec 09 21:31:37 domain systemd[1]: [/etc/systemd/system/listmonk.service:67] Unknown lvalue 'LockPersonality' in sectio...ervice'
Dec 09 21:31:37 domain systemd[1]: [/etc/systemd/system/listmonk.service:71] Unknown lvalue 'ProtectKernelModules' in s...ervice'
Dec 09 21:31:37 domain systemd[1]: [/etc/systemd/system/listmonk.service:74] Unknown lvalue 'PrivateUsers' in section 'Service'
Hint: Some lines were ellipsized, use -l to show in full.
```

Is it possible to send an email notification when that happens? I know my CSF firewall is able to email me about events. I modified these lines to try to fix the issue:
I commented out the 2nd line here:
```
EnvironmentFile=-/etc/default/listmonk
# EnvironmentFile=-/etc/default/listmonk-%i
```
I'm not sure whether to comment out these lines:
```
StateDirectory=listmonk-%i
Environment=HOME=%S/listmonk-%i
WorkingDirectory=%S/listmonk-%i
```
Or change them to:
```
StateDirectory=/etc/listmonk
Environment=HOME=/etc/listmonk
WorkingDirectory=/usr/bin
```
I was thinking to [simplify it the same way bamboowonder did](https://github.com/knadh/listmonk/issues/843#issuecomment-1159822295), but that guide doesn't cover creating a new user for `User=<yourUser>`. I saw another page showing that you can run it without specifying a user, so for now I've done that.

However, as noted here https://github.com/knadh/listmonk/issues/843#issuecomment-1836023524 another way to "fix this" is by running `systemctl restart listmonk` after `systemctl daemon-reload`. Is there a way to make it do that automatically?

Oh, I think it didn't restart after the server rebooted. So this needs to be changed too:
`Restart=on-failure` to `Restart=always` or something else: https://www.freedesktop.org/software/systemd/man/latest/systemd.service.html

## Summary:

This file needs major changes to be more compatible. I suggested in the issue-thread that we could have two versions or two sets of commands in the same file with one set commented out by default.

Perhaps we could leave the current file mostly the same _(and/or rename it to `listmonk-multi@.service`)_ and create a new `listmonk-simple@.service` file containing:
```
[Unit]
Description=listmonk email service
ConditionPathExists=/etc/listmonk/config.toml
Wants=network.target
# The PostgreSQL database may not be on the same host but if it
# is listmonk should wait for it to start up.
After=postgresql.service

[Service]
Type=simple
PermissionsStartOnly=true
WorkingDirectory=/usr/bin
EnvironmentFile=-/etc/default/listmonk
ExecStartPre=/usr/bin/mkdir -p "/listmonk/uploads"
ExecStartPre=/usr/bin/listmonk --config /etc/listmonk/config.toml --upgrade --yes
ExecStart=/usr/bin/listmonk --config /etc/listmonk/config.toml $SYSTEMD_LISTMONK_ARGS
Restart=always

# Use systemd’s ability to disable security-sensitive features
# that listmonk does not explicitly need.
# NoNewPrivileges should be enabled by DynamicUser=yes but systemd-analyze
# still recommended to explicitly enable it.
NoNewPrivileges=True
# listmonk doesn’t need any capabilities as defined by the linux kernel
# see: https://man7.org/linux/man-pages/man7/capabilities.7.html
CapabilityBoundingSet=
# listmonk only executes native code with no need for any other ABIs.
SystemCallArchitectures=native

# Make /home/, /root/, and /run/user/ inaccessible.
# If you set ExecStartPre=/usr/bin/mkdir -p "listmonk/uploads" to a directory in /home/ or /root/ it will cause uploads to fail 
# See https://github.com/knadh/listmonk/issues/843#issuecomment-1836023524
ProtectHome=True

# Make sure files created by listmonk are only readable by itself and
# others in the listmonk system group.
UMask=0027
# listmonk only needs to support the IPv4 and IPv6 address families.
RestrictAddressFamilies=AF_INET AF_INET6

[Install]
WantedBy=multi-user.target
```

Maybe `ExecStartPre=/usr/bin/mkdir -p "/listmonk/uploads"` should be changed to `/etc/listmonk/uploads`.

And still missing is a static directory. https://listmonk.app/docs/templating/#system-templates

MaximilianKohler:
@kmohrf does everything look good now with this service and [the new simple.service](https://github.com/knadh/listmonk/pull/1622)? The only thing I'm not sure about is the `HOME` in `Environment=HOME=%S/listmonk-%i`, and if I broke something by changing `ExecStartPre=/usr/bin/mkdir -p "${HOME}/uploads"`.

MaximilianKohler:
## Logs:

I'm not sure which is best, but when reviewing https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html I saw these options and added them commented out in the simple.service https://github.com/knadh/listmonk/blob/master/listmonk-simple.service file:
```
#LogsDirectory=/etc/listmonk
#CacheDirectory=/etc/listmonk/cache
```
I assume that listmonk doesn't use a cache, so I can probably remove that? I don't know how to get listmonk to save logs with that though, and this (alternative?) may be the easiest/best way (which is working for me):
```
ExecStart=/bin/bash -ce "exec /usr/bin/listmonk --config /etc/listmonk/config.toml --static-dir /etc/listmonk/static >>/etc/listmonk/listmonk.log 2>&1"
```
So I added it as an option. I'm not sure if it conflicts with anything else.

MaximilianKohler:
I switched to AlmaLinux 9 and added back many of [the commands I previously removed due to them causing error messages](https://github.com/knadh/listmonk/issues/843#issuecomment-1836023524) because of an old systemd version.

```
systemctl --version
systemd 252
```

Running these consecutively:
```
systemctl daemon-reload
systemctl status listmonk
```
Still results in:
```
/etc/systemd/system/listmonk.service:54: Invalid device node path 'False', ignoring.
```
Not sure if that's the expected result though. Line 54 is this:
```
# listmonk doesn’t handle any specific device nodes.
DeviceAllow=False
```

listmonk@.service
```diff
@@ -8,12 +11,23 @@ After=postgresql.service
 
 [Service]
 Type=simple
+PermissionsStartOnly=true
 EnvironmentFile=-/etc/default/listmonk
 EnvironmentFile=-/etc/default/listmonk-%i
-ExecStartPre=/usr/bin/mkdir -p "${HOME}/uploads"
+ExecStartPre=/usr/bin/mkdir -p "/etc/listmonk/uploads"
```
knadh, line 17: This shouldn't default to `/etc/listmonk` as a) the app should be run as a non-root user and hence, permission to write to /etc/ is unlikely b) arbitrary user files/content should not go in `/etc/*`.

MaximilianKohler: Where should they go, given that `${HOME}` is incompatible with `ProtectHome=True`?

knadh: Did a bit of Googling and looks like `BindPaths, ReadWritePaths` are the way to go. I am not familiar with it, so will have to experiment.

PR #1622, opened 2023-12-11
Create listmonk-simple.service

MaximilianKohler:
Alternative to some of the changes in https://github.com/knadh/listmonk/pull/1614

### DIFF
--- a/listmonk-simple.service
+++ b/listmonk-simple.service
@@ -31,6 +31,7 @@ RestartSec=5
 #StateDirectory=/etc/listmonk
 #LogsDirectory=/etc/listmonk
 #CacheDirectory=/etc/listmonk/cache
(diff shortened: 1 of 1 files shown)


MaximilianKohler:
I was going to readd `DynamicUser=True` _(since it didn't seem to give an error)_ but I saw that it was introduced in systemd 232 https://0pointer.net/blog/dynamic-users-with-systemd.html, and CentOS 7 is only on systemd 219.
```
# Create dynamic user for listmonk service instance, and folders it can write to
# https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#DynamicUser=
DynamicUser=True
StateDirectory=/etc/listmonk
LogsDirectory=/etc/listmonk
CacheDirectory=/etc/listmonk/cache
```

PR #1659, opened 2024-01-09
Fix #1657 - Prevent adding attachments multiple times in campaign test messages

### LINKED ISSUE #1657: Attachments get multiplied when sending a campaign test message to multiple subscribers
**Version:**
 - listmonk: v2.5.1
 - OS: Ubuntu

When sending a test message for a campaign with attachments to multiple subscribers, the attachments are added to the e-mails multiple times. 

With each additional subscriber, the file gets attached one more time (the first subscriber in the list gets one attachment, 2nd gets two, and so on).

If the campaign has multiple attachments, all of them are duplicated.

Only tested this with the built-in `email` messenger so far. Can confirm it **doesn't** happen when sending the actual campaign (Start Campaign).

**Steps to reproduce**
1. Create a campaign with attachment(s)
2. Send a test message to at least 2 subscribers at once
3. Check the received e-mails. Subscriber 2 will receive all the attachments twice

**Screenshots:**
Test message with 1 attachment (cover.png) sent to 2 subscribers
Sending test message

Emails received by first (left) and second (right) subscriber
Received emails

mpilnan:
Create a copy of the campaign object before sending test message to each subscriber. This way the attachments are not shared between subscribers.

Fixes #1657

### DIFF
--- a/cmd/campaigns.go
+++ b/cmd/campaigns.go
@@ -453,7 +453,8 @@ func handleTestCampaign(c echo.Context) error {
 	// Send the test messages.
 	for _, s := range subs {
 		sub := s
-		if err := sendTestMessage(sub, &camp, app); err != nil {
+		c := camp
+		if err := sendTestMessage(sub, &c, app); err != nil {
 			app.log.Printf("error sending test message: %v", err)
 			return echo.NewHTTPError(http.StatusInternalServerError,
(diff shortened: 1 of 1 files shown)


knadh:
Ah, good spot! This is the right way to fix it.

PR #1758, opened 2024-02-25
fix: query campaign does not search numbers

### LINKED ISSUE #1713: Unable to search campaigns if the campaign name has numbers.
**Version:**
 - listmonk:  v3.0.0
 - OS: Ubuntu 20

**Description**
Unable to search a campaign with numbers in the name. 

**Screenshots:**
Screenshot 2024-02-05 111319
Screenshot 2024-02-05 111334

scmmishra:
This PR fix a problem with the, where previously it couldn't handle alphanumeric strings effectively, for instance "Testing1234". 

## The problem

The problem arises becuase of the limitations encountered with PostgreSQL's full-text search when dealing with alphanumeric strings. The core of the issue lies in the process of tokenization and indexing is optimized for natural language text and not for patterns that include numbers.

Full-text search breaks down text into tokens based on spaces and punctuation, then converts these tokens into lexemes. Alphanumeric strings might not be broken down or analyzed in a way that's useful for matching partial strings, especially when numbers are involved.

This can be easily demostrated with the following statements 

```sql
$ SELECT to_tsvector('Testing') @@ to_tsquery('Testing123');
FALSE
```
For the above instance the `tsvectors` for `Testing` is `'test':1`, however the tsquery is `'testing123'`, which is a bad match

```sql
$ SELECT to_tsvector('123') @@ to_tsquery('Testing123');
FALSE
```
Similarly, the vectors for `123` is `'123':1` and the tsquery is `testing123`

## The solution

To fix this, we can add a fallback, which does substring matching using the `ILIKE` operator to ensure accurate results for alphanumeric queries.

Fixes #1713

### DIFF
--- a/queries.sql
+++ b/queries.sql
@@ -534,7 +534,7 @@ FROM campaigns c
 WHERE ($1 = 0 OR id = $1)
     AND (CARDINALITY($2::campaign_status[]) = 0 OR status = ANY($2))
     AND (CARDINALITY($3::VARCHAR(100)[]) = 0 OR $3 <@ tags)
-    AND ($4 = '' OR TO_TSVECTOR(CONCAT(name, ' ', subject)) @@ TO_TSQUERY($4))
+    AND ($4 = '' OR TO_TSVECTOR(CONCAT(name, ' ', subject)) @@ TO_TSQUERY($4) OR CONCAT(c.name, ' ', c.subject) ILIKE $4)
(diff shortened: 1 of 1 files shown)


MaximilianKohler:
Any idea why this is working in 2.5.1 but not 3.0?

scmmishra:
> Any idea why it works in 2.5.1 but not 3.0?

I am assuming this is the commit https://github.com/knadh/listmonk/commit/8e3e1b9af860f10879dda2339f5ca9204d1241db that changed it from an ILIKE to a TSVector query

PR #1843, opened 2024-04-25
Add ability to match SMTP servers to email addresses

StratusFearMe21:
Hi! I've just installed Listmonk onto my server and I'm loving it so far. My use case for it is starting an email list within my high school, and your software allows me to do that within the school's closed email system. Which means that our school emails can only send and receive emails from other school emails.

I'd love to be able to open up my mailing list to personal emails as well, but as I've said, my school's emails can't send to anyone else outside the school email system, so I'm opening this PR to allow Listmonk to match subscribers to SMTP servers. That way emails to school accounts are sent through the school SMTP server, and emails to personal accounts are routed elsewhere

20240425_13h55m04s_grim

Thanks in advance for considering my PR!

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
-    "dev": "vite --host 0.0.0.0",
+    "dev": "vite",
     "build": "vite build",
     "serve": "vite preview",
     "lint": "eslint --ext .js,.vue --ignore-path .gitignore src",
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
-    "dev": "vite",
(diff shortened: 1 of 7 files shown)


knadh:
Hi @StratusFearMe21. This is an interesting usecase, but I think it's a very rare and niche one to be a feature in the core. A more robust approach is to setup a relay/SMTP proxy that receives all e-mails and then proxies it upstream based on arbitrary conditions.

StratusFearMe21:
> A more robust approach is to setup a relay/SMTP proxy that receives all e-mails and then proxies it upstream based on arbitrary conditions.

I see, do you know of any SMTP proxies that do this? I tried finding one before opening this PR but was unsuccessful

knadh:
I haven't done this in a long while, but I remember Postfix "transport maps" being able to route based on different rules.

https://www.linuxbabe.com/mail-server/postfix-transport-map-relay-map-flexible-email-delivery

Issue #1770, opened 2024-03-06
Campaign planified => Attachment disappear


ArabianCat:
**Version:**
 - listmonk: 3.0.0
 - OS: Ubuntu 22.04

**Description of the bug and steps to reproduce:**
When I planned an email with an attachment, the attachment disappear few minutes agos. So when the email has been send, the receiver don't have the attachment.
Can u fix it ?

ArabianCat:
Hello,
I specify I try many times with many compaign, and the problem is same.
Cordially

knadh:
I am unable to reproduce this. Tried the following:
- Create a campaign, add attachment and save. Switch content to plaintext and then send. Works fine.
- Create campaign, switch to plaintext, add attachments and send. Works fine.

ArabianCat:
Hello.
The problem appears during campaign planning, not in creation and send of the campaign.

MaximilianKohler:
@ArabianCat you should list your steps.

ArabianCat:
Ok I'll list : 
- First, I create the campaign, i select a template, add a text and add a attachment
- Secondly, I plan my campaign (for exemple, tommorow at 9 AM

And few minute laters, the attachment dispair in the planned campaign.

knadh:
I'm unable to reproduce this.
- Create campaign.
- Enter text.
- Add attachments.
- Set schedule.
- Save

Attachments don't disappear.

ArabianCat:
Hello @knadh ,
I make a campaign planned yesterday with an attachment.
Today, the attachment dispair.
I think after 24h of schedule, the attachment dispair (Last time it was a campaign planned for a week later).

knadh:
Couldn't reproduce this. Created a new campaign, added attachments, scheduled it to be sent after X minutes. Came back and the campaign was sent, attachments were delivered and were still visible on the campaign.

There is no logic in listmonk that removes attachments from a campaign automatically.

ArabianCat:
The problem is not resolved.
Maybe it's not logical but it's a true problem.
All the attachment disparair after approximatly 24 hours on all campaign (send or not) and it's a true problem.

ArabianCat:
I test yesterday, and i've the problem, so the problem is real.

knadh:
Could you share screenshots of the steps you’re following? How the campaign looks when it is created and what happens after the schedule triggers it.

ArabianCat:
[...]

PR #1894, opened 2024-06-12
fix vanishing attachments in campaign in more the one campaigns setup

zevszym:
On update sql statements need delete only defined ids not all rows in the table.
solve: https://github.com/knadh/listmonk/issues/1770

### DIFF
--- a/queries.sql
+++ b/queries.sql
@@ -827,7 +827,7 @@ clists AS (
 ),
 med AS (
     DELETE FROM campaign_media WHERE campaign_id = $1
-    AND media_id IS NULL or NOT(media_id = ANY($19)) RETURNING media_id
+    AND media_id IS NULL or media_id = ANY($19) RETURNING media_id
 ),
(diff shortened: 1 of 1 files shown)


queries.sql
```diff
@@ -827,7 +827,7 @@ clists AS (
 ),
 med AS (
     DELETE FROM campaign_media WHERE campaign_id = $1
-    AND media_id IS NULL or NOT(media_id = ANY($19)) RETURNING media_id
+    AND media_id IS NULL or media_id = ANY($19) RETURNING media_id
```
knadh, line 830: This is not correct, right? Media elements that are NOT in the given list should be deleted.

zevszym: Thank you for your explanation.  Of Course you have the right. After your tip I saw that problem here is caused by  SQL logic operator precedence. I made new pull request: https://github.com/knadh/listmonk/pull/1936

Issue #1807, opened 2024-04-03
bounce.go:170: error processing SES notification: notification type is not bounce


robguthrie:
**Version:**
 - listmonk: [eg: v3.0.0]
 - OS: [e.g. Ubuntu / Docker]

**Description of the bug and steps to reproduce:**
I've setup SNS notification topic as described in https://github.com/knadh/listmonk/issues/1668#issue-2080438101 but I'm not getting bounce notifications, in the logs I can see this error:
```
bounce.go:170: error processing SES notification: notification type is not bounce
```

I've tripple checked my settings and it seems correct.

knadh:
SES sends many types of notifications. Perhaps you've enabled non-bounce notifications. I'm not familiar with SES config though.

robguthrie:
Sorry, I should have added these are logs in response to me using the bounce@simulator.amazonses.com address to test. There are no other logs happening during this time, and the count of logs corresponds to my tests sent. 
The SNS topic confirmation request works correctly though.

Is there a way to enable more log output so I can see the body of the request?

knadh:
There are no verbose payload logs. It should be fine to ignore those errors as non-bounce (bounce/complaint) notifications are discarded with that particular message. The simulator must be sending non bounce/complaint notifications.

MaximilianKohler:
I wouldn't say this is fine and to ignore it. I haven't tested v3.0, but earlier versions had this issue before being fixed. I'm on v2.5.1 with that exact setup and everything works normally.

If you're only emailing bounce@simulator.amazonses.com, then there's clearly something wrong. The SES simulator does not send non-bounce/complaint notifications.

Since there are no verbose logs, you may want/have to try v2.5.1 to test, and if you're still getting the error then you've misconfigured something. If it works properly then it's a bug with v3.0.

Make sure that in this part of the instructions the 3rd option (1 bounce, 2 complaint, 3) is left blank:

> In the AWS console, go to [Simple Email Service](https://console.aws.amazon.com/ses/) and click "Verified identities" in the left sidebar.
> Click your verified domain and go to the "Notifications" tab.
> Next to "Feedback notifications", click "Edit".
> For "Bounce feedback" and "Complaint feedback", select the SNS topic you created earlier.

MaximilianKohler:
>Is it safe to rollback from v3 to v2.5.1 without changing the database?

According to the docs it is not safe. I'd recommend cloning your server temporarily to test it.

benstaf:
[...]

PR #1907, opened 2024-06-21
docs: add important step regarding ses in bounces.md

bhartshorn:
Bounces seemingly aren't sent if this setting is only configured for the domain, needs to be configured for send-as email addresses also.

Per https://github.com/knadh/listmonk/issues/1807#issuecomment-2151855416

I made this change in my environment recently and will test it shortly.


bhartshorn:
Just a heads up that this change did indeed work for my environment, bounces are now handled correctly.

PR #1963, opened 2024-07-28
Fixes #1957: Ensure unique S3 object keys by adding numeric suffix.

### LINKED ISSUE #1957: Unique media name in S3
**Version:**
 - listmonk: [eg: v3.0.0]
 - OS: [e.g. Ubuntu - docker version]

**Description of the bug**
Hello,
When a new media is uploaded, a new id is created in database and the file is uploaded on S3.
However, if the media filename already exists on S3, the existing file is simply overwrited and we have two records in database, pointing on the same file.

For that reason, on client side, I created a check method that removes all media entries in database that have the given target filename before uploading a new one. For testing this method : 
- I created 200 entries with same filename, using `POST /api/media` (no problem)
- I loop over media id to remove all the entries using `DELETE /api/media` endpoint

Now here is the bug: after exactly 99 media entries removed successfully, I get the following HTTP 500  for the 100th :
```
{'message': 'Error creating Media: sql: no rows in result set'}
```

If I retry with new deletion loop, it goes well up to  100 entries again.

**How you can reproduce**

Using [listmonk-api](https://github.com/Knuckles-Team/listmonk-api) python client :
```python

# Listmonk server
username = "user"
password = "password"
listmonk_api_url = "http://localhost:9000"
client = listmonk_api.Api(url=listmonk_api_url, username=username, password=password)

# Creating 200 media entries for the same file
attachment = "/path/to/testfile.txt"
media_ids = []
for i in range(200):
     media_infos = client.upload_media(file=attachement)
     media_ids.append(["data"]["id"])

# Now deleting the created entries
for media_id in media_ids:
    response = client.delete_media(media_id)
    if not response.get("data", False):                       # <- Fails exatly after 100 deletions
        print(f"Fail to delete media {idx} : {media_id}")
        print(response)
        exit(1)
exit(0)
```

abhinavxd:
- Removed unused functions and variables in filesystem.go

### DIFF
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -39,8 +39,7 @@ func makeFilename(fName string) string {
 func appendSuffixToFilename(filename, suffix string) string {
 	ext := filepath.Ext(filename)
 	name := strings.TrimSuffix(filename, ext)
-	newName := fmt.Sprintf("%s_%s%s", name, suffix, ext)
-	return newName
+	return fmt.Sprintf("%s_%s%s", name, suffix, ext)
 }
 
 // makeMsgTpl takes a page title, heading, and message and returns
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -35,6 +35,14 @@ func makeFilename(fName string) string {
 	return filepath.Base(name)
 }
 
+// appendSuffixToFilename adds a string suffix to the filename while keeping the file extension.
+func appendSuffixToFilename(filename, suffix string) string {
+	ext := filepath.Ext(filename)
+	name := strings.TrimSuffix(filename, ext)
+	newName := fmt.Sprintf("%s_%s%s", name, suffix, ext)
+	return newName
+}
+
 // makeMsgTpl takes a page title, heading, and message and returns
 // a msgTpl that can be rendered as an HTML view. This is used for
 // rendering arbitrary HTML views with error and success messages.
--- a/internal/media/providers/s3/s3.go
+++ b/internal/media/providers/s3/s3.go
@@ -64,16 +64,15 @@ func NewS3Store(opt Opt) (media.Store, error) {
 
 // Put takes in the filename, the content type and file object itself and uploads to S3.
 func (c *Client) Put(name string, cType string, file io.ReadSeeker) (string, error) {
-	// Paths inside the bucket should not start with /.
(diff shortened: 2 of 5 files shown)


knadh:
Thanks for the PR!

Thinking about it, rather than each provider implementing its own logic, it'd be simpler to do it centrally in `handleUploadMedia()` and check in the database. Roughly:

- Add a `GetMediaByFilename(fname, provider string)` function to `core/media.go`
- This will require a new query to be added to queries.sql also
- Use this in `handleUploadMedia()` to assert the incoming filename and tweak it if it already exists.
- Remove the existing assertion logic from the filesystem provider.

abhinavxd:
Yep, makes sense. I will update the PR.

abhinavxd:
image
I added the DB level check for unique filename, but it fails when multiple files with the same name are selected for upload.

The previous provider level check handles this properly.

image

abhinavxd:
@knadh Changes made as discussed.

cmd/utils.go
```diff
@@ -35,6 +35,14 @@ func makeFilename(fName string) string {
 	return filepath.Base(name)
 }
 
+// appendSuffixToFilename adds a string suffix to the filename while keeping the file extension.
+func appendSuffixToFilename(filename, suffix string) string {
+	ext := filepath.Ext(filename)
+	name := strings.TrimSuffix(filename, ext)
+	newName := fmt.Sprintf("%s_%s%s", name, suffix, ext)
```
knadh, line 42: Can return from here directly with the variable `newName`.

internal/media/providers/s3/s3.go
```diff
@@ -64,15 +64,16 @@ func NewS3Store(opt Opt) (media.Store, error) {
...
+
 	// Upload input parameters
 	p := simples3.UploadInput{
 		Bucket:      c.opts.Bucket,
 		ContentType: cType,
 		FileName:    name,
 		Body:        file,
-
-		// Paths inside the bucket should not start with /.
-		ObjectKey: c.makeBucketPath(name),
```
knadh, line 75: This doesn't need to change.

knadh (changes requested):
Couple minor things to fix.

abhinavxd:
Made the changes requested.

PR #1964, opened 2024-07-29
Update developer-setup.md

Bowrna:
I have updated the doc with default username and password, as I didn't know that when I initially setup and have to find it from config.toml I hope it's ok to add this as part of documentation. If otherwise, please let me know the reason.


knadh:
This isn't correct. When you run `./listmonk --new-config`, it generates a new sample config file with a random password. If you're manually copying the sample config, you've to edit it to set the database params to begin with, at which point you'll end up setting the password anyway.

Bowrna:
@knadh I have been following [this](https://listmonk.app/docs/developer-setup/) page for setting up the env and here it is mentioned to copy the config rather than generating one. Yes, I edited to set the DB param and didn't look at the default username/password when editing the config file. That's why I had to come back to the config file to look at the username and password and hence added the same in docs too.

Please let me know your thoughts on this.

knadh:
Closing this as in the upcoming v4, there is no default username/password. Those keys are gone from the config completely.

Issue #1879, opened 2024-05-30
Cant clone campaign where lists no longer exist


AshPowell:
Not sure if bug or feature enhancement - but if a list no longer exists, it stops you being able to clone a campaign they're linked to.

Bowrna:
could i try my hands on this issue as part of FOSS Hack?

knadh:
@Bowrna yes, sure.

Bowrna:
This shows an error message saying "There are no subscribers in the selected list to create the campaign" when we try to clone the campaign associated with deleted lists.
But at the backend, a campaign is getting created and on refresh at front end, you can see a copy of the campaign without any associated list. 
Attaching screenshots for reference
Taking a clone of the campaign with deleted list, show following message
<img width="1150" alt="Screenshot 2024-07-29 at 2 30 36 PM" src="https://github.com/user-attachments/assets/e796f60d-fb34-42ac-8b8d-7870fbf57034">
On page reload, it shows the following clone 
<img width="1150" alt="Screenshot 2024-07-29 at 2 26 03 PM" src="https://github.com/user-attachments/assets/44caf38b-da39-488d-9915-56721d0c5fd1">

PR #1966, opened 2024-07-31
issue fix for clone campaign with empty list

Bowrna:
fixes: #1879

This issue occurred as the CTE query returned the id from campaign after inserting entry into campaign_lists table. But in this case, there is no insert happening and the result set returned was empty. Fixed the query to return id irrespective of insert happening in the campaign_lists table. 

Thanks,
Bowrna


knadh:
Thanks. I'll review and merge this week.

Issue #1717, opened 2024-02-06
error sending message '[Title]': remote error: tls: user canceled


kulterryan:
**Version:**
 - listmonk: [eg: v1.0.0]
 - OS:Ubuntu/Railway

**Description of the bug and steps to reproduce:**
Getting TLS error after a specific time interval when logged in the application.

**Screenshots:**
![image](https://github.com/knadh/listmonk/assets/36154235/fdf7a0f7-2235-4e16-a175-6c587c475a11)

knadh:
hm, sounds like an an SMTP/TLS configuration issue. What SMTP provider are you using?

kulterryan:
I'm using AWS SES

knadh:
Unable to reproduce this with SES in production. Could've something to do with network configuration.

MaximilianKohler:
I just started getting this with SES as well on v2.5.1. I've been sending 2-3k emails per day without issue and gradually increased it to 50k, and it happened for the first time now even though my SES limit is 500k.

```
2024/03/21 08:52:00 start processing campaign (<campaign name> 2024-03-21 50k)
2024/03/21 09:12:01 error sending message in campaign <campaign name> 2024-03-21 50k: subscriber 968723: 421 Connection closed by server. Too many messages in one session
2024/03/21 09:12:01 error sending message in campaign <campaign name> 2024-03-21 50k: subscriber 968726: remote  error: tls: user canceled
2024/03/21 09:12:01 error sending message in campaign <campaign name> 2024-03-21 50k: subscriber 968724: 421 Connection closed by server. Too many messages in one session
2024/03/21 09:12:01 error sending message in campaign <campaign name> 2024-03-21 50k: subscriber 968725: 421 Connection closed by server. Too many messages in one session
2024/03/21 09:12:01 error sending message in campaign <campaign name> 2024-03-21 50k: subscriber 968728: 421 Connection closed by server. Too many messages in one session
2024/03/21 09:12:01 error sending message in campaign <campaign name> 2024-03-21 50k: subscriber 968727: 421 Connection closed by server. Too many messages in one session
2024/03/21 09:12:01 error sending message in campaign <campaign name> 2024-03-21 50k: subscriber 968729: remote  error: tls: user canceled
2024/03/21 09:12:01 error sending message in campaign <campaign name> 2024-03-21 50k: subscriber 968730: 421 Connection closed by server. Too many messages in one session
[...]

Issue #172, opened 2020-08-08
Campaign progress issue


heinrichreimer:
In v0.6.2-alpha, the campaign progress seems to be broken.
It shows the number of Mails fetched from the DB and inserted to the message queue.
In my opinion it should show the number of messages that were successfully processed and sent from the queue.
That leads to situations where the progress is like 1234/1234 (100%), but the campaign is still running.
Also it is hard to monitor progress, as database connections are usually much faster than sending messages.

knadh:
It's not broken, but is a trade-off. Each send cycle involves fetching records from the DB with an offset marker which is recorded in the campaigns table. The campaign stats uses this metric to show an approximation of progress. Up until v0.6x, this batch size was hardcoded in the core (5k I think), so the progress was based on blocks of incremental 5k fetches. In the upcoming release, this number is configurable. By setting a lower number, the progress accuracy can increase.

About keeping track of live e-mail counts, it's a little tricky. Where as the current progress simply uses existing campaign offsets from the DB with no additional state management, counting sends will require addition states. Hopefully this'll improve at some point, but until then, the batch size control should help.

heinrichreimer:
Okay, I'll try out some different values for the batch size. Thanks for this great app!

PR #1967, opened 2024-08-01
Update configuration.md +performance, batch size

MaximilianKohler:
I don't know if this is correct, but there's no info in the docs about these kinds of settings. All I could find was one comment here https://github.com/knadh/listmonk/issues/172 with some info.

I'm trying to figure out a solution to https://github.com/knadh/listmonk/issues/1717. I'm hoping that the larger the batch size, the longer the delay between retries will be, but I'm not sure if that's true.

I saw another "performance" section https://listmonk.app/docs/maintenance/performance/, but it's under "maintenance" so this seems more appropriate under "configuration".

### DIFF
--- a/docs/docs/content/configuration.md
+++ b/docs/docs/content/configuration.md
@@ -139,3 +139,12 @@ with any Timezone listed [here](https://en.wikipedia.org/wiki/List_of_tz_databas
 ## SMTP Blocked Ports
 Some server hosts block SMTP ports (25, 465) so you have to get request to unblock them i.e. [Hetzner](https://docs.hetzner.com/cloud/servers/faq/#why-can-i-not-send-any-mails-from-my-server).
 
+## Performance
+
+### Batch size
+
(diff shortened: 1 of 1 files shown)


knadh:
> The campaign stats use this metric to show an approximation of progress. By setting a lower number, the progress accuracy can increase.

This is in correct. As of last version, every message is tracked granularly showing accurate progress. Batch size has no effect on the rate calculation.

> Under `Settings -> SMTP -> retries`, listmonk will attempt to retry any failed deliveries after the full batch has been sent. So increasing the batch size will increase the time between retries. This can be helpful if you're getting 4xx sending errors.

This is also not correct. When a message is sent, if it fails, it's retried on a new connection then and there, before being marked as an error (if the retry also fails). There is no batching in retries.

MaximilianKohler:
So `batch size` had a minor role in previous versions but is largely meaningless now?

knadh:
It comes in handy when working with a large number of subscribers. Pulling bigger batches (which'll use more memory, but reduce the number of DB queries) will help achieve more throughput.

PR #1976, opened 2024-08-06
adding lists docs for public lists api

Bowrna:
Added missed-out documentation at lists.md for public lists api.


knadh:
This isn't just to retrieve public lists, but it's a public unauthenticated API for scripting subscription forms etc. from public endpoints.

Bowrna:
@knadh I missed that. let me remove the authentication part from curl command and update the API doc

Bowrna:
@knadh I have updated this PR. Let me know if this is ok, or does the public API's have to go to a different place in documentation.

PR #1978, opened 2024-08-07
adding docs for bounces api

Bowrna:
Adding docs for bounces API


MaximilianKohler:
FYI, it is not in the left-side nav menu https://listmonk.app/docs/apis/bounces/. If you added other pages, it might be the case for them too.

PR #1983, opened 2024-08-08
Update querying-and-segmentation.md - Querying subscribers who viewed the campaign email

MaximilianKohler:
I'm not sure if this is the correct place for this. Additionally, I wasn't sure if `campaign_views` should be added to the "Database fields" section at the top. 

Additionally, when I go to `/admin/subscribers` -> advanced search, and paste in `subscribers.email LIKE 'John%' AND status = 'blocklisted'`, I get an error notification: `Error fetching Subscribers: pq: column reference "status" is ambiguous`
status error

Also, that error notification does not get logged, and disappears too fast. I think the time it stays visible should be increased, but I'm not sure how to do that.

I am trying to add these various commands to the docs: https://gist.github.com/MaximilianKohler/e5158fcfe6de80a9069926a67afcae11#postgresql

### DIFF
--- a/docs/docs/content/querying-and-segmentation.md
+++ b/docs/docs/content/querying-and-segmentation.md
@@ -66,6 +66,13 @@ subscribers.email LIKE 'John%'
 subscribers.email LIKE 'John%' AND status = 'blocklisted'
 ```
 
+#### Querying subscribers who viewed the campaign email
+
+```sql
+-- Find all subscribers who viewed the campaign email.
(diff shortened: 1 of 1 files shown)


knadh:
We can add a `## Sample snippets` section on that page and list this and other commonly required snippets there.

MaximilianKohler:
There is a `Sample SQL query expressions` section already, which I put it under. Not sure how that's different from `sample snippets`. https://listmonk.app/docs/querying-and-segmentation/

I think @Bowrna added many of the snippets recently, but maybe not the ones on this page.

PR #1998, opened 2024-08-15
Possible to unsubscribe from lists as well as from campaigns, even if the emails are transactional. 

macgayverarmini:
(text using translator).
Dear all, we have software that serves several companies, and when customers of these companies make in-person purchases, the system sends the company's customers an email with the payment slip and the invoice required by the government. However, it happens that sometimes, because it is an in-person purchase, the person does not have an email address and ends up passing on the email addresses of relatives or friends, especially older people, so we need to use the liskmonk list system, and it should be possible to unsubscribe from lists as well as from campaigns, even if the emails are transactional. We made a temporary change to the code to allow this, but I would like to leave the suggestion here.

### DIFF
--- a/cmd/handlers.go
+++ b/cmd/handlers.go
@@ -196,6 +196,11 @@ func initHTTPHandlers(e *echo.Echo, app *App) {
 		"subUUID"))
 	e.POST("/subscription/wipe/:subUUID", validateUUID(subscriberExists(handleWipeSubscriberData),
 		"subUUID"))
+
+	// one-click unsubscribe from list
+	e.POST("/unsubscribe/:subUUID/:listUUID", unsubscribeList)
+	e.GET("/unsubscribe/:subUUID/:listUUID", noIndex(validateUUID(subscriberExists(handleListSubscriptionPage), "subUUID", "listUUID")))
+
 	e.GET("/link/:linkUUID/:campUUID/:subUUID", noIndex(validateUUID(handleLinkRedirect,
 		"linkUUID", "campUUID", "subUUID")))
 	e.GET("/campaign/:campUUID/:subUUID", noIndex(validateUUID(handleViewCampaignMessage,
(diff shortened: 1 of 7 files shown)


knadh:
Thanks for the PR @macgayverarmini. I'll only be able to look into after the v4.0.0  (upcoming) is released.

waza-ari:
Hi, please allow me to ask a question - how exactly does one unsubscribe from a transactional mail? There is no list of users maintained or mapped to that transactional mail, could you please share any detail?

macgayverarmini:
> Hi, please allow me to ask a question - how exactly does one unsubscribe from a transactional mail? There is no list of users maintained or mapped to that transactional mail, could you please share any detail?

Hi, Imagine this: we are a software company. Several other companies use our software to manage their businesses, such as accounts receivable, accounts payable, cash flow, generating and sending invoices to their customers, and issuing sales tax documents, etc. These companies generally don't have their own corporate email, and when one of these companies does have an email with their domain, it's usually low-quality and practically goes straight to spam. Another problem is small-scale providers that block SMTP ports, preventing our customers from sending emails from within our system. Naturally, they call us for support instead of their internet provider.

Due to all these reasons, we are adopting Amazon SES, along with Listmonk on our server. Our software will consume the Listmonk API to create lists for each company (unique UUID per list) for specific actions. For example, when Company A sends a sales tax document, it will have a list with its UUID and a prefix for this type of document. The template is standard for all users, and if the recipient eventually doesn't authorize or no longer wants to receive these emails from that company, they can unsubscribe on their own. By working this way, we will allow all our customers who use our ERP software to send emails by default, without our support team having to configure Gmail SMTP (or the API that the company owner might accidentally remove authorization for on their Android phone).

From this point on, the emails are not campaigns, as they are triggered because the person who received them actually made the purchase. However, to avoid abuse of our Amazon SES, we are adopting this transactional email subscription process.

MaximilianKohler:
This should be reopened right?

Issue #1521, opened 2023-09-12
Allow for CORS Origin fixation or bypass


duraki:
It would be good to have some kind of CORS settings from within app, for example spoofing/bypassing based on user-provided CORS value. A lot of times you will see no support for header settings from reverse proxy, therefore this would come handy for all users.

I remember Ackee Analytics [implemented](https://docs.ackee.electerious.com/#/docs/CORS%20headers#platforms-as-a-service-configuration) optional `.env` key which allows user to configure accepted origins.

knadh:
Please see this issue that was just closed. https://github.com/knadh/listmonk/issues/1523

I think it's best for this to be in the reverse proxy and not in listmonk's core for now.

duraki:
> A lot of times you will see no support for header settings from reverse proxy

what .. have you read what I've even wrote 🥲

candidexmedia:
Good news: https://github.com/knadh/listmonk/pull/2000#issuecomment-3446420585

duraki:
@candidexmedia thanks for the headups! Was fucking time fosho.

knadh:
CORS 'allowed origins' can now be configured in Settings -> Security in [v6.0.0](https://github.com/knadh/listmonk/releases/tag/v6.0.0).

Issue #1523, opened 2023-09-12
CORS issue, Listmonk configured with Nginx reverse proxy


krmjitsingh:
I have configure listmonk with nginx reverse, getting cores error this issue is only with react with nodejs everything working fine

Access to fetch at 'https://listmonk.unikaksha.com/api/subscribers' from origin 'https://localhost:3000/' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled
![CORS](https://github.com/knadh/listmonk/assets/96057208/c0ed07f4-229a-4297-87d3-6e87b3d51c18)

henk23:
Same issue here. Nginx configuration has the correct headers:

    add_header Access-Control-Allow-Origin $allow_origin always;
    add_header Access-Control-Allow-Methods 'GET, HEAD, POST, PUT, PATCH, DELETE, OPTIONS' always;
    add_header Access-Control-Allow-Credentials true always;
    add_header Access-Control-Allow-Headers 'Authorization, Content-Type' always;

But the OPTIONS request fails

```
❯ curl 'https://listmonk.my-website.com/api/public/subscription' -X OPTIONS
{"message":"Unauthorized"}
❯ curl 'https:/listmonk.my-website.com/api/public/subscription' -X OPTIONS -I
HTTP/1.1 401 Unauthorized
Server: nginx/1.18.0 (Ubuntu)
Date: Thu, 14 Sep 2023 08:57:37 GMT
Content-Type: application/json; charset=UTF-8
Content-Length: 27
Connection: keep-alive
Www-Authenticate: basic realm=Restricted
Access-Control-Allow-Methods: GET, HEAD, POST, PUT, PATCH, DELETE, OPTIONS
Access-Control-Allow-Credentials: true
Access-Control-Allow-Headers: Authorization, Content-Type
```

duraki:
@krmjitsingh this will help you #1521 - just apply my patch on the fork and build listmonk again

henk23:
Hey, sorry @duraki I have no idea what you're referring to, there seems to be a history that I don't know.

Just saying: Configuring the reverse proxy correctly will lead to Listmonk throwing a 401 at an OPTIONS request, that should be a 2xx. Right?

duraki:
[...]

PR #2000, opened 2024-08-16
Add optional CORS header middleware

bgervan:
Optional CORS header middleware with configs like:

[cors]
enabled = true
allowed_origins = ["example.com"]

or with environment variable like (in docker compose for example):

```
  app:
    <<: *app-defaults
    container_name: listmonk_app
    command: [ sh, -c, "yes | ./listmonk --config config.toml" ]
    depends_on:
      - db
    volumes:
      - ./config.toml:/listmonk/config.toml
    environment:
      - LISTMONK_cors__enabled=true
      - LISTMONK_cors__allowed_origins__0=*
      - LISTMONK_cors__allowed_origins__1=http://localhost:3000
      - LISTMONK_cors__allowed_origins__2=google.com
      - LISTMONK_cors__allowed_origins__2=another.com
```
 Note: the google.com is overriden with another.com in this example.
 
 With enabled false, everything works as before, so it doesn't break backward the compatibility. Without array of origins, it shows a warning, but works like cors disabled. 
 
 Inspiration for this PR: https://github.com/knadh/listmonk/issues/1521

### DIFF
--- a/cmd/handlers.go
+++ b/cmd/handlers.go
@@ -3,13 +3,12 @@ package main
 import (
 	"bytes"
 	"crypto/subtle"
-	"net/http"
-	"path"
-	"regexp"
-
 	"github.com/knadh/paginator"
 	"github.com/labstack/echo/v4"
 	"github.com/labstack/echo/v4/middleware"
+	"net/http"
+	"path"
+	"regexp"
 )
 
 const (
@@ -63,6 +62,23 @@ func initHTTPHandlers(e *echo.Echo, app *App) {
 			app.log.Println(err.Error())
 		}
 		e.DefaultHTTPErrorHandler(err, c)
+	}
+
+	// CORS middleware
+	if ko.Bool("cors.enabled") {
+		var allowedOrigins = ko.Strings("cors.allowed_origins")
+		// If the allowed origins are not set, the middleware will not be used.
+		if len(allowedOrigins) > 0 {
+			e.Use(middleware.CORSWithConfig(middleware.CORSConfig{
+				AllowOrigins: ko.Strings("cors.allowed_origins"),
+				AllowHeaders: []string{echo.HeaderOrigin, echo.HeaderContentType, echo.HeaderAccept},
+			}))
+			app.log.Println("CORS middleware enabled")
+			app.log.Printf("Allowed origins: %v", allowedOrigins)
+		} else {
+			app.log.Println("CORS middleware not enabled as allowed origins are not set properly")
+		}
+
 	}
 
 	// Admin JS app views.
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -7,6 +7,8 @@ import (
 	"log"
 	"os"
 	"os/signal"
+	"sort"
+	"strconv"
 	"strings"
 	"sync"
 	"syscall"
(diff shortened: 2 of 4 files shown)


knadh:
Thanks for the PR @bgervan, but I don't think it's ideal for CORS configuration to be in listmonk. CORS, TLS and any other HTTP configuration is best handled in the reverse proxy that you would be using to expose listmonk to the internet, like Nginx or Caddy.

Please see my comment on the old issue: https://github.com/knadh/listmonk/issues/1523#issuecomment-1724910581

bgervan:
> Thanks for the PR @bgervan, but I don't think it's ideal for CORS configuration to be in listmonk. CORS, TLS and any other HTTP configuration is best handled in the reverse proxy that you would be using to expose listmonk to the internet, like Nginx or Caddy.
> 
> Please see my comment on the old issue: https://github.com/knadh/listmonk/issues/1523#issuecomment-1724910581

I saw that, but it's very common to handle this on application side (in django for example).
Also on caprover the same issue mentions that it's better to handle on app side not in the nginx config. That comment made me create it optional.

In pikapod, it blocks the usage as they are not supporting cors configuration.

knadh:
Django is a framework for building web applications, so it's natural for it to have configuration for a wide variety of HTTP related things. In fact, it is very uncommon for "consumer" web applications like listmonk to have CORS configuration.

listmonk does have built in SSL/TLS either and fully relies on a reverse proxy to provide it. Also, except for a couple of `/api/public/*` handlers, every other endpoint is private and meant to be used only with authentication.

> that it's better to handle on app side not in the nginx config

That would be incorrect. A reverse proxy like Nginx is the best place to handle config like TLS, CORS etc. transparently without having every web application build support for those.

bgervan:
It still gives extra complexity to run nginx beside listmonk. Local docker compose too, but in case of cloudflare tunnel the ssl is handled by cloudflare, but for cors only I would need to tun a reverse proxy as well and point my tunnel to that.

Pikapod is a cheap way to run it without needing to manage, but with custom integrated form, we cannot use it.

candidexmedia:
> Pikapod is a cheap way to run it without needing to manage, but with custom integrated form, we cannot use it.

Is this also an issue with Railway.app?

bgervan:
> > Pikapod is a cheap way to run it without needing to manage, but with custom integrated form, we cannot use it.
> 
> Is this also an issue with Railway.app?

I am not familiar much with Railway, but if you use the docker image for the deployment without an nginx, yes.
 
https://docs.railway.app/guides/create

candidexmedia:
> I am not familiar much with Railway, but if you use the docker image for the deployment without an nginx, yes.
> 
> https://docs.railway.app/guides/create

Looks like the latest Railway template has CORS supporr via Caddy: https://github.com/railwayapp-templates/listmonk/commit/39d3e5349fbc8c3ecfa800e8e276b393b3b66b0d#commitcomment-148663870

ericwaetke:
I really question this decision honestly …
Especially since the solution is right here!

This cors issue renders the entire API pretty much useless. I mean, I cant even simply spin up the docker-compose locally and build my application that talks to the API because of cors.

At the _very_ least there should be a disclaimer in the main readme above the docker setup, that you cant actually use the app fully without a proxy.
Or even better, have a docker compose with a preconfigured proxy, so I can develop locally and don’t have to search through month-old issue and pull requests.

Or even better: just handle CORS in the App itself :D

ericwaetke:
Additionally, this here: https://github.com/knadh/listmonk/issues/1521

I’m don’t know/think what you’re suggesting is even possible with a proxy like traefik. If it is, it’s above my paygrade – but as I can’t simply change what proxy we use, this is starting to become a real showstopper.

And that’s really a shame, as I love the rest of the app!

knadh:
@ericwaetke which API endpoint are you trying to access via the browser?

ericwaetke:
/api/subscribers

I want to basically have a form on my website where people can join the mailing list.
And ideally I’d like to statically build my website, which kind of forces me to send this request from the client.

I got a very basic caddy docker-compose setup ready though, but having multiple proxies behind each other is not my ideal hosting solution 😅

knadh:
CORS can now be setup directly from the admin UI. Added this after increasing user demand and complains of UX-pain.

https://github.com/knadh/listmonk/commit/cdf0a5c153f4967e477bad318840726d1172184f

duraki:
Well it was fucking time @knadh? Thanks for closing up #1521 (and resolving the issue) 🤷‍♂️ Too bad most of the gents moved out to other OSS self-hosting solutions by the time you rounded up this highly requested issue. Eh ... Cheers 🍻

samuelsogbesan:
Hi, is there any documentation on how to configure this in the admin UI? I can't find the setting

samuelsogbesan:
I'm guessing this also also coming in the next release

candidexmedia:
@knadh Silly question, but what's the difference between the new CORS setting

image

and the env variable `LISTMONK_ORIGIN_0`?

Are they the same thing? Should I remove those variables and replace them with the setting? Should I be using both?

knadh:
Hi @candidexmedia. I'm not sure where `LISTMONK_ORIGIN_0` is from. listmonk doesn't have any env var configuration for CORS, only the new feature in Settings.

candidexmedia:
> I'm not sure where LISTMONK_ORIGIN_0 is from.

It's from here (Railway's Listmonk template):

https://github.com/railwayapp-templates/listmonk/commit/39d3e5349fbc8c3ecfa800e8e276b393b3b66b0d#comments

https://github.com/railwayapp-templates/listmonk/blob/main/Caddyfile

PR #2034, opened 2024-09-06
nit:  English

bish0polis:
'software', when used as a noun, is not a 'countable' type, and it does not get an article like 'a'.  It's like 'traffic'.


knadh:
Indeed!

Issue #1841, opened 2024-04-24
Selecting to blocklist a selection, blocklists the whole list


AshPowell:
**Version:**
 - listmonk: 3.0
 - OS: Ubuntu

**Description of the bug and steps to reproduce:**
I'm trying to blocklist subscribers who have unsubscribed from a list. So:

1) Select 'unsubscribed' number on a list to go into the unsubscribed users view.
2) Filter subscribers.status = 'enabled'
3) Select top box to select all in view, then click 'Select all 123' next to 20 subscribers selected.
4) Click blocklist and confirm to blocklist 123 subscribers
5) Whole list of subscribers get blocklisted, not the selected query.

The first time this happened I thought I must have selected the whole list by accident, second time I am 100% sure this is a bug, I 100% confirmed I clicked only 123 subscribers and confirmed on the blocklist popup to blocklist 123 subscribers.

Thanks!

Bowrna:
can I take a look into this issue?

knadh:
Yes please. Thank you!

Issue #1985, opened 2024-08-10
Bulk actions not working in the List detail view


tanmay-predisai:
**Version:**
 - listmonk: v3.0.0
 - OS: Ubuntu

**Description of the bug and steps to reproduce:**
1. Click on Menu -> List option
2. Click on the unsubscribe number CTA in the list to open this URL https://<domain>/admin/subscribers/lists/1?subscription_status=unsubscribed
3. Click Export. The Confirmation popup will state the exact number of subscribers in the popup.
4. When the actual export is downloaded, it has the entire list instead of the stipulated number of subscribers.

5. Click on the Select All CTA. The blacklist CTA will be visible now.
6. Click on Blacklist - The Confirmation popup will state the exact number of subscribers in the popup. 
7. Once you confirm, it will actually Blacklist the entire list instead of only doing it to the selected contacts. 

Also, what is the difference between Blacklist and unsubscribed contacts?

**Screenshots:**

MaximilianKohler:
>Also, what is the difference between Blacklist and unsubscribed contacts?

Unsubscribe is only from the one list. Blocklist is from all lists, including future ones.

PR #2041, opened 2024-09-12
passed the subscription_status param

Bowrna:
fixes: #1841
related_to: #1985

This issue existed as the subscription_status filter was not passed to the backend. Added a fix by including a new param to send the status, and altered the numbering in parameterized query.

The query affected is a base query template. So all the queries using this base query were in turn affected.


Bowrna:
Let me know if I had to add anything more to this PR. Thanks!

knadh:
Thanks for the PR @Bowrna!

PR #2064, opened 2024-10-03
Fix typo in German translation

j-maas:
Just a small typo I noticed.


knadh:
Hi @j-maas. `E-mail` (lowercase `m`) isn't a typo though. Ref: https://en.wikipedia.org/wiki/Email#Terminology

j-maas:
@knadh In German, I'm used to the spelling "E-Mail".

Another observation: Most of the instances in that file are already spelled "E-Mail". The places where I changed them were inconsistent with that.

j-maas:
Thanks for the quick response!

By the way, I was really frustrated with phpList and tried out Listmonk via PikaPods today and it went great! I'm pretty sure it'll be easy to convince the others to switch.

So thanks for the great software!

Issue #2058, opened 2024-09-29
Markdown editor lacks syntax highlighting


olib14:
Simply, the the _Markdown_ editor does not have syntax highlighting. The _Raw HTML_ editor already has it, and _Rich text_/_Plain text_ do not require it.

![image](https://github.com/user-attachments/assets/5b6b87e6-88ed-4f73-aa86-2aa780eece41)

The right window, KWrite, supports syntax highlighting for a variety of markdown features, such as raw HTML.

olib14:
Additionally, it would be good if the editor supported breaking things whilst retaining indentation - this helps to structure the text and make editing more pleasant:

![image](https://github.com/user-attachments/assets/a411a8c0-b1b1-43ce-8fc9-0a8c75ce4b49)

The "Product" section is all a single line with one level of indentation (two spaces, as four denotes code blocks). In the current markdown editor, the text spills with no consideration for indentation.

This is also quite useful for bullet-point lists.

knadh:
@Bowrna would you like to explore this? listmonk uses [@kazzkiq/CodeFlask](https://github.com/kazzkiq/CodeFlask) for HTML syntax highlighting. It has support form Markdown as well I think.

Bowrna:
yes I will pick this one after raising PR for this https://github.com/knadh/listmonk/issues/1728

PR #2068, opened 2024-10-04
markdown syntax highlighter

Bowrna:
related: #2058 

Added fix to support markdown syntax highlighter. As code-flask doesn't support markdown by default, I had to import prism-markdown. Added styles in shadow DOM as style from prismjs doesn't percolate/get injected into shadow DOM.

Let me know if anything has to be modified/ fixed in this PR.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -18,6 +18,7 @@
     "codeflask": "^1.4.1",
     "dayjs": "^1.11.10",
     "indent.js": "^0.3.5",
+    "prismjs":"^1.29.0",
     "qs": "^6.10.1",
     "textversionjs": "^1.1.3",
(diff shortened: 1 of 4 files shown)


knadh:
Looks fine @Bowrna. The JS lint test is failing though. vscode (or any other editor with a VueJS plugin) should ideally have auto-formatted it.

knadh:
Tested this locally. One minor issue. The editor box styling is a bit broken. It's missing borders and it's height is off. Should be the same as other editors.

image

Bowrna:
Ah I missed noticing it. Let me fix it.

Bowrna:
@knadh updated the styling sheet for markdown-editor.

Issue #2107, opened 2024-10-28
Docker Upgrade Error


mrhll:
**Version:**
 - listmonk: Latest
 - OS: Docker/Debian 12

Just followed the documentation to upgrade docker :
docker compose down app
docker compose pull
docker compose up app -d

There was a 500 Server Error and docker service keeps restarting forever.

knadh:
You've to run the `--upgrade` [step](https://github.com/knadh/listmonk/releases/tag/v4.0.1) if you're using the old Docker compose file.

```shell
docker-compose down
docker-compose pull && docker-compose run --rm app ./listmonk --upgrade
docker-compose up -d app db
```

mrhll:
That worked. Thank you !
Please mind updating the upgrade instructions as it was not clear enough.

PR #2108, opened 2024-10-28
Update upgrade.md

MaximilianKohler:
Solves https://github.com/knadh/listmonk/issues/2107


knadh:
For the new Docker compose file, the manual `--upgrade` step is not necessary. Simply doing a down/pull/up will auto-upgrade the DB.

PR #2113, opened 2024-10-28
Add the upgrade command to the Docker steps

JamesChevalier:
Copy the supporting text from the Binary upgrade steps


JamesChevalier:
This was addressed in a direct commit: https://github.com/knadh/listmonk/commit/998b6e300af688b27ef02434313d0fc67447771b

Issue #2069, opened 2024-10-05
Email Campaign performance settings -  paused with "timed out waiting for free conn in pool"


tanmay-predisai:
**Version:**
 - listmonk: v3.0.0
 - OS: Ubuntu 22.04.3

**Campaign stopped abruptly after sending 120K emails with notification 
Status paused
Sent	0 / 0
Reason	Too many errors **

When I checked the logs, there were >450 entries of "error sending message in campaign <campaign name> : subscriber <#>: timed out waiting for free conn in pool". I assume some threshold got hit and the campaign paused. 

We have been successfully sending campaigns with >1M subscribers till now. However, this is the second time this week that the campaign has stopped. No settings have changed - only the number of subscribers were reduced in this campaign. 

We are using AWS SES with enough limits (daily email sending limits >1.5M, per second email sending limit = 500) 

**Screenshots:**
Screenshot of the Performance Settings page - 
<img width="914" alt="Screenshot 2024-10-05 at 10 07 04 AM" src="https://github.com/user-attachments/assets/d262a9fb-38c9-4161-9c64-7a42540de0d2">

Screenshot of the SMTP settings page - 
<img width="881" alt="Screenshot 2024-10-05 at 10 08 05 AM" src="https://github.com/user-attachments/assets/264fcdc2-c503-4105-8ef7-449b9fdb5725">

The settings have not been changed recently. 

How do we debug this / understand what has gone wrong and fix it? Thank you for your kind help!

knadh:
Concurrency = 100, which means you'll have 100 concurrent messages being attempted to send. But you've set max messages to 2 per second, which makes this large number moot anyway.

But, your SMTP pool's max conns is set to 10, which means at max, you can only send 10 messages concurrently. The timeout is 5s, which means each connection can wait up to 5s. Any attempt to send more will wait on the pool and then just throw a timeout if the connections (SES) have slowed down for any reason.

You should reduce concurrency to something like 5-10 (given you've sent max messages to 2) and increase SMTP max conns to match.

The exact combination of these vary greatly based on many factors (speed of SMTP, number of messages to be send, the size of each message etc.), so you've to experiment and figure out.

tanmay-predisai:
Hello - Thank you for your reply!

We tried with: 
Max. connections = 200
Max retries = 5
Concurreny = 100
max messages = 2

[...]

PR #2133, opened 2024-11-01
Update en.json - SMTP Max. connections info

MaximilianKohler:
Addresses https://github.com/knadh/listmonk/issues/2069


i18n/en.json
```diff
@@ -426,7 +426,7 @@
     "settings.mailserver.idleTimeout": "Idle timeout",
     "settings.mailserver.idleTimeoutHelp": "Time to wait for new activity on a connection before closing it and removing it from the pool (s for second, m for minute).",
     "settings.mailserver.maxConns": "Max. connections",
-    "settings.mailserver.maxConnsHelp": "Maximum concurrent connections to the server.",
+    "settings.mailserver.maxConnsHelp": "Maximum concurrent connections to the server. Should = Message rate x Concurrency.",
```
knadh, line 429: This isn't correct. Max conns can be anything and don't need to be message_rate x concurrency.

MaximilianKohler: I based it on your comment here https://github.com/knadh/listmonk/issues/2069#issuecomment-2394927415

>your SMTP pool's max conns is set to 10, which means at max, you can only send 10 messages concurrently
>
>You should reduce concurrency to something like 5-10 (given you've sent max messages to 2) and increase SMTP max conns to match.

Would it be more accurate to say `SMTP Max. connections should = the Performance -> Concurrency number`?

MaximilianKohler: Your last comment in that issue also makes it sound like this current help text is incorrect:

>If concurrency = 10 and message_rate = 10, then up to 10x10=100 messages may be pushed out every second

Per your comment, it sounds like "Message rate" is the actual limit, and "Concurrency" just enables it.

So I think the help text for each of the performance settings needs some tweaking/clarification.

knadh: I think we should change the setting's label to `Max message rate` instead of `Message rate`.

MaximilianKohler: How would you be getting 50k/minute with 

> Concurrency: 250
> Message rate: 50

50/s x 60 seconds = 3k/minute.

knadh: In the instance of the settings I shared (which keep getting tweaked based on campaigns) 250 workers with each worker doing a max of 50/sec = 12.5k/sec max. For larger campaigns, we spawn and add multiple SMTP servers and raise the limits accordingly.

hm, sending 20/sec with 10 in settings, I'll have to investigate that.

MaximilianKohler: >250 workers with each worker doing a max of 50/sec = 12.5k/sec max

Ok, I guess there's a communication issue because this indicates that the `If concurrency = 10 and message_rate = 10, then up to 10x10=100 messages may be pushed out every second` help text is correct, even though [it sounded like you contradicted that recently](https://github.com/knadh/listmonk/issues/2069#issuecomment-2455468817).

Yet [my own experiments](https://github.com/knadh/listmonk/issues/2069#issuecomment-2452911380) indicate that help text is wrong.

* You're sending at 50,000 per minute
* 12500 per second is 750,000 per minute

None of this is making sense.

knadh: Lost track. After reviewing this, the help text is correct. It begins by saying `Max..`. So it's the maximum upper limit permitted, but it does not mean that the program can actually push out so many e-mails to an SMTP server. There are many bottlenecks (CPU/network resources, database bottlenecks, network, and of course, the capacity of the SMTP server itself).

We've set an upper limit of 250 workers with 50/worker/sec max. With that, including all the bottlenecks (primarily the capacity of the SMTP servers to handle messages), we're achieving a net throughput of 50k. If the SMTP servers aren't performant, it can drop to any number. It doesn't mean that the system will be achieve a 750k/min throughput.

Basically, like the help text indicates, it is the `Maximum` upper limit, not a guarantee of any throughput.

Issue #2072, opened 2024-10-05
Replace code flask with a more active alternative


sweetppro:
Code class was last updated 5yrs ago, and can be slow on large inputs. 

https://github.com/WebCoder49/code-input seems to be quite active with the last release only 3 months ago. The setup looks similar to code flask and it also utilises prism.js or highlight.js for syntax highlighting.

knadh:
Was thinking about this the other day. CC @Bowrna, since you've a CodeFlask related PR in the works.

Bowrna:
Yes I can check this.

Bowrna:
hello @Remceau i was working on this PR https://github.com/knadh/listmonk/pull/2139 that replaces code flask with code-input. Due to health issues, I didn't try closing the PR as promised. I will try closing this.

knadh:
Ended up replacing CodeFlask with [CodeMirror](https://codemirror.net/).

Ref: https://github.com/knadh/listmonk/commit/ba5e77b868370ba8c0d263b845583b739771e2ea

sweetppro:
Looking forward to a new beta with this patch applied 😊

sweetppro:
@knadh thanks, Im running v5.beta.2 and CodeMirror is much more performant and much easier to add custom styles

PR #2139, opened 2024-11-05
replacing codeflask with code-input

Bowrna:
related: #2072 

This is work in progress to replace codeflask with code-input for showing syntax highlighted text/code editor inside campaign content.


Bowrna:
I have some css related issue in this PR. Let me try debugging and closing this issue before this weekends.

Bowrna:
Could this PR be reviewed?
If there is any feedback, let me know.

knadh:
Hi @Bowrna. There seems to be several issues with this.

The HTML editor is completely broken. This is what I'm getting:
image

Referring `./node_modules/*` in the frontend HTML file is not correct. All modules should be bundled in the JS compilation.

Bowrna:
@knadh Let me fix this and update here. I missed this.

frontend/src/components/HTMLEditor.vue
```diff
@@ -13,66 +21,60 @@ export default {
...
+        // window.codeInput.setDefaultTemplate('syntax-highlighted');
+      }
     },
   },
 
   mounted() {
-    this.initHTMLEditor(this.$props.value || '');
-  },
-
-  watch: {
```
Et-008, line 70: Why is the below code being used only in markdownEditor and not in HTMLeditor?

```
watch: {
    value(newVal) {
      if (newVal !== this.data) {
        this.data = newVal;
      }
    },
  },
```

Bowrna: Let me check this one @Et-008

knadh:
@Bowrna, this is now resolved in https://github.com/knadh/listmonk/commit/ba5e77b868370ba8c0d263b845583b739771e2ea.

Last week, I spent some time looking at CodeInput, and not only is it tedious to integrate, but has a bunch of deal-breaker open issues. I went ahead and integrated CodeMirror (which is a much larger lib), which is battle tested and was actually a breeze to integrate.

Thanks again for the PR.

Bowrna:
Thanks for this update @knadh 
I had a tough time with this task. Could you share the deal-breaker open issues that you faced regarding CodeInput? Integration was tedious, and while I faced multiple issues, I was unsure if I could call out the issues, and I kept trying to fix them. If you could share about these deal-breaker issues, it would be helpful for me to understand the thought behind this decision, and I can learn to navigate through such scenarios.

knadh:
CodeInput is unfortunately not packaged in clean, importable, reusable way. Or packaged at all.

- You can't yarn/npm `install` it and `import` and use it in the app directly. The only way seems to be to add build steps in the Makefile to copy the bundles from the `./node_modules` directory into the `frontend/dist` directory. This is not nice.
- Since it's not `import`able, there's no way to integrate it seamlessly. When the editor view renders, have to do DOM manipulation to dynamically inject a `<script>` tag pointing to the dist, wait for it to load, and then init. Or, have a dedicated page that does all of this and `<iframe>` it in the editor, then do hacky iframe manipulation. Also not nice.
- Plugins seem to be outright non-installable via yarn/npm. The repo has them committed, but they don't seem to be packaged.
- Multiple fundamental [open issues](https://github.com/WebCoder49/code-input/issues) including broken RTL, cursor placement, text alignment, and scroll issues.

Too many bad trade-offs. The better trade-off was to just bite the bullet and use CodeMirror, a much bigger, but cleaner, and robust library.

CodeMirror integration is seamless. Install the necessary modules, and simply import, compose, and use them.
https://github.com/knadh/listmonk/blob/feat-visual-editor/frontend/src/components/CodeEditor.vue

Issue #2186, opened 2024-11-29
Unable to select any messenger except "email" in v4.0


centja1:
**Version:**
 - listmonk: 4.0
 - OS: Ubuntu

**Description of the bug and steps to reproduce:**

1. Fresh install of listmonk v3.0.0, create a webhook messenger
2. Go to the "Test Campaign", you're able to select the newly created messenger
3. Upgrade to v4.0.0
4. Go to the "Test Campaign", only "email" and two blank entries are in the messenger select box

Screenshots of a docker container running the code in the forthcoming PR are also supplied

![Listmonk v3.0.0 Install](https://github.com/user-attachments/assets/39d01e69-2b7a-4f88-997e-57efec9c6204)
![Listmonk v3.0.0 Create Webhook](https://github.com/user-attachments/assets/fadb6dce-261a-4ba4-b245-af3da188973d)
![Listmonk v3.0.0 Test Campaign](https://github.com/user-attachments/assets/9ae32923-0cf5-408c-8b5b-875c2bb38cf7)
![Listmonk v3.0.0 Select Webhook for Campaign](https://github.com/user-attachments/assets/89b6f98f-a934-436f-822f-678ead6534f6)
![Listmonk v4.1.0 Upgrade](https://github.com/user-attachments/assets/dc99f6b7-c9b1-4c92-ba13-6e1a2fdcc176)
![Listmonk v4.1.0 Verify webhook configuration](https://github.com/user-attachments/assets/ec754e81-b14e-42da-8585-d03c71cea4c1)
![Listmonk v4.1.0 Misconfigured messenger select](https://github.com/user-attachments/assets/43cfe09a-cbaa-4c82-b342-51ef205e1bf0)
![Listmonk v4.1.0 Config API](https://github.com/user-attachments/assets/5ff019d3-dbef-4163-a07e-bc3a7bf37a43)

![Modified code for PR](https://github.com/user-attachments/assets/4e47aef2-81c2-4027-894a-cee022dcfbc3)

knadh:
This is fixed in #2187.

PR #2187, opened 2024-11-29
Corrected loading of messengers from serverConfig on line 629

centja1:
Starting with v4.0, it was impossible to select any messenger besides email for a campaign due to an issue loading the messenger list from the API on line 629 of Campaign.vue.

The change was making m.name be just m since the messengers are listed as an array of strings on /api/config

### DIFF
--- a/frontend/src/views/Campaign.vue
+++ b/frontend/src/views/Campaign.vue
@@ -626,7 +626,7 @@ export default Vue.extend({
     },
 
     messengers() {
-      return ['email', ...this.serverConfig.messengers.map((m) => m.name)];
+      return [...this.serverConfig.messengers.map((m) => m.name)];
     },
   },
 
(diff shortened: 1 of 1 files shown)


centja1:
#2186

frontend/src/views/Campaign.vue
```diff
@@ -626,7 +626,7 @@ export default Vue.extend({
     },
 
     messengers() {
-      return ['email', ...this.serverConfig.messengers.map((m) => m.name)];
+      return [...this.serverConfig.messengers.map((m) => m.name)];
```
knadh, line 629: This doesn't work. Should be:

```javascript
return [...this.serverConfig.messengers];
```

centja1: My mistake, I committed the incorrect code.  I've updated it and resubmitted

Issue #2231, opened 2025-01-02
Configure Sliding window per campaign


pankajmyt:
**Is your feature request related to a problem? Please describe.**
Currently a sliding window of max X ops per duration can be specified on global level. However there are certain campaigns in which the send rate has to be distributed over different range of hours/days. 
like email we want to send over 3 days, sms we want to send within 1 hour.

If I have to adjust the sliding window value, have to pause all running campaigns. This will help so that only 1 campaign is paused.

**Describe the solution you'd like**
Have a setting available on campaign management page which allows to configure number of ops in a sliding window.

pankajmyt:
Duplicate of #2043 closing this.

pankajmyt:
Ticket mentions #2043 a variable window, however what is required is on per-campaign level.

pankajmyt:
This is working for us so closing this.

PR #2232, opened 2025-01-02
Move sliding window to configurable per campaign

pankajsoni19:
This moves message sliding window from global to per campaign. 
Updates to db migration, UI, and text localisation.


pankajsoni19:
* message batch size should be defined on the campaign level. [ maybe will include later ]
* smtp error threshold to move to per smtp config.  [ maybe will include later ]

pankajmyt:
Ticket #2231

knadh:
Thanks for the PR @pankajmyt. Also, sliding window per campaign is a niche feature. If anything, SMTP per campaign (which has been requested multiple times), should come first. I've been planning that for the upcoming release.

Also, unfortunately, this has a large number of changes including new API handlers, new campaign fields etc. I've no bandwidth to review this currently.

Since this feature is not in the pipeline, I'm closing the PR. Perhaps this can re-looked at in the future.

Thanks again.

pankajsoni19:
yes, its quite a bit :) 
I am customising it to our need, I have added per campaign smtp config, list cloning, fixing subscriber drop on campaign pause, and long running campaigns.

PR #2279, opened 2025-02-01
Update Node.js version to 18 in app.Dockerfile

savioribeiro:
This update changes the Node.js version in app.Dockerfile from 16 to 18 to resolve compatibility issues with eslint-define-config@2.1.0.

### DIFF
--- a/dev/app.Dockerfile
+++ b/dev/app.Dockerfile
@@ -1,6 +1,6 @@
 FROM golang:1.20 AS go
 
-FROM node:16 AS node
+FROM node:18 AS node
 
 COPY --from=go /usr/local/go /usr/local/go
 ENV GOPATH /go


knadh:
Could you please confirm that the switch to Node 18 doesn't break any JS dependencies? Also, I'm unsure what the issue with eslint-define-config is.

savioribeiro:
I'm not sure if any dependencies broke with node 18, although I was able to run it today using version 16 and 18. Anyway, I managed to recover the log with the warnings and errors I initially had:

```
warning Resolution field "jackspeak@2.1.1" is incompatible with requested version "jackspeak@^3.1.2"
[2/4] Fetching packages...
warning chart.js@4.4.7: The engine "pnpm" appears to be invalid.
error eslint-define-config@2.1.0: The engine "node" is incompatible with this module. Expected version ">=18.0.0". Got "16.20.2"
warning eslint-define-config@2.1.0: The engine "pnpm" appears to be invalid.
error Found incompatible module.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
make: *** [Makefile:37: frontend/node_modules] Error 1
```

PR #2280, opened 2025-02-01
Fix: LogView regex for milliseconds & add span file support

savioribeiro:
This PR enhances the log parsing functionality by updating the regex to support milliseconds ([#30be235](https://github.com/knadh/listmonk/commit/30be235e2ab87697d563f0adc8fef8d72a8f3ce6)) in LogView.vue. Additionally, the .file class was merged with .timestamp in styles.scss, ensuring consistent styling and allowing users to select both elements together with proper spacing.

### DIFF
--- a/frontend/src/assets/style.scss
+++ b/frontend/src/assets/style.scss
@@ -973,11 +973,12 @@ section.analytics {
       margin-bottom: 2px;
     }
 
-    .timestamp {
+    .timestamp, .file {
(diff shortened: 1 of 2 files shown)


knadh:
Thanks for the PR. Could you share a screenshot showing the changes on the UI?

savioribeiro:
@knadh 

Before this PR:

before

After:

after

PR #2299, opened 2025-02-10
Add support to Jumpcloud OIDC

amantinetti:
**Motivation**

Support Jumpcloud

**Modifications**

Support "email_verified" field as string "true"

Now the application can receive  a string or boolean value ​​in the email_verified field from the OIDC

### DIFF
--- a/internal/auth/auth.go
+++ b/internal/auth/auth.go
@@ -33,11 +33,25 @@ const (
 	sessTypeOIDC   = "oidc"
 )
 
+type StringOrBoolean bool
+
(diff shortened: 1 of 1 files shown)


knadh:
Thanks for the PR @amantinetti, but it's not ideal to add provider-specific hacks to OIDC. It's unfortunate that JumpCloud doesn't conform to the [standard](https://openid.net/specs/openid-connect-core-1_0.html#rfc.section.5.1) and instead of a boolean, uses the string: `"1"`. There could be providers that use arbitrary indicators like `"y"`, `"true"`, `"yes"` etc.

Issue #2419, opened 2025-04-19
Exporting Individual Views and Individual Clicks as CSV


destrex271:
**Describe the solution you'd like**
In our forked version of listmonk, we have added the functionality to export individual views and clicks. I believe this feature is something that people might like. Although alternatives like directly querying PostgreSQL and fetching the data are good but for someone who is just operating the interface without any knowledge of SQL it is a difficult task to perform.

The current implementation looks something like this.

![Image](https://github.com/user-attachments/assets/33662b19-66d8-44a6-b4db-e31f72963e4c)

In case of links on exporting we get the following CSV file

```csv
campaignId,url,name,email
"14","https://google.com","Akshat","akshatdev2711@gmail.com"
```

PR #2420, opened 2025-04-19
Export Individual Analytics as CSV

destrex271:
Hi!
This PR is for the following issue:
 -  #2419 

Feedback is welcome !

### DIFF
--- a/cmd/campaigns.go
+++ b/cmd/campaigns.go
@@ -513,6 +513,115 @@ func (a *App) GetCampaignViewAnalytics(c echo.Context) error {
 		return err
 	}
 
+
+	return c.JSON(http.StatusOK, okResp{out})
+}
+
+func handleGetCampaignIndividualViews(c echo.Context) error{
+	var (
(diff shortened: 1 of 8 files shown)


knadh:
Thanks for the PR @destrex271. While this is useful in some scenarios, I think it's a bit too niche to be included in the core.

destrex271:
> Thanks for the PR @destrex271. While this is useful in some scenarios, I think it's a bit too niche to be included in the core.

Ohk Understood! 

Although I would like to suggest an alternative: Maybe we can introduce a similar functionality to export all the analytics data(that we currently have in listmonk) in the maintenance section? This can be useful if someone wants to export all their data, whether it be analytics or the subscribers, to some third party platform without wrestling with the database.

knadh:
@destrex271 it would make sense to add an `Export` button the analytics page I guess.

destrex271:
> @destrex271 it would make sense to add an `Export` button the analytics page I guess. Closing this PR for now.

Cool! I'll get started on it

Issue #2395, opened 2025-04-09
Multi-Tenancy Permissions


xeruf:
**Is your feature request related to a problem? Please describe.**
I would like a user on my instance to create and manage their own mailing lists - it is no problem if I see them.

**Describe the solution you'd like**
A lists:create permission?
Maybe for proper permissions a list role needs to be created automatically then and assigned to the user.

For testing, I created a user only with subscribers:import permission, and the user can see Lists>Forms but get an error when clicking on it.
![Image](https://github.com/user-attachments/assets/ed2b551a-47c8-4abb-b607-5508fee33205)

I am running v4.1.0 via docker compose

xeruf:
Plus, after adding the lists:get_all permission, I can see a list of lists, but the interface gets stuck when selecting a list:

![Image](https://github.com/user-attachments/assets/47e40d60-8d45-4cd6-985a-9fdb52997d12)

knadh:
@xeruf that shouldn't happen! Could you share the URL where this is happening and the exact steps to reproduce this issue?

xeruf:
1. Create a user with only subscribers:import permission and login
2. Click on Lists>Forms
3. Get permission denied settings:get error (Error 1)
4. Add lists:get_all permission and reload
5. Click on Lists and then on a List - I am guessing it is refusing because of the missing lists:manage_all permission, but there is no error message, just console errors and no response from the interface (Error 2)

Generally I find the permission system needs quite some work - a lot of permissions are subsets of each other but treated as independent - you cannot manage a list without having the permission to view it?
So maybe set permissions with a Slider? (Feature Request 1)
For example:

Lists Access: <-- Only List Roles -- View All -- Manage All -->

This also makes it more clear that the lists permissions are on top of the List Roles, which was not clear to me in the beginning - I thought lists permissions are needed fro any lists access

dragonwocky:
@knadh @xeruf I think I've got the same use case: users should have scoped access to a set of lists, but be able to add new lists to that set themselves.

So far the closest I've gotten is a list role + user role with `lists:manage_all` enabled and `lists:get_all` disabled. This lets users manage only lists added to the list role, and create new lists - there's just no way for them to then manage those new lists. Ideally it should be possible to add lists created by a user with a list role to that list role automatically?

[...]

PR #2451, opened 2025-05-02
Add `list:create` permission for multi-tenancy

dragonwocky:
See #2395.

Separating out list creation from list management permissions allows for proper scoping of list roles, such that users can create and modify as many lists as they want without other users being able to get a look at their lists or subscribers.

When a list is created by a user with a list role, that list is automatically added to their list role. If the user has no list role, then the list is created as normal. A new list role could be created for these users, but I don't think it's necessary. I assume that users will either have a list role, or have the `lists:get_all` permission and not need a list role to see their created list.

(As a bonus fix while I was working on this, I've removed the advanced search field from the UI for users without the `subscribers:sql_query` permission.)

I've included a database migration with this to add the new permission to existing users. Hope this is helpful and can get merged! Thanks for all your work on this project, of everything I've tried it's stood out best by far.


knadh:
Thanks for the PR @dragonwocky.

> When a list is created by a user with a list role, that list is automatically added to their list role.

The issue with this is that all users who have the same list role will end up getting access to the list.

dragonwocky:
Isn't that the idea of list roles, though? If two users have the same list role currently there's no way for them not to see all the same lists. If you wanted users not to have the same scope of lists, you would put them in different list roles. This works the way I'd want/expect for my use case.


I could add a warning/explanation to the list role menus of what to expect and/or make the feature toggle-able for individual list roles?

xeruf:
that really makes sense and would help us a lot!
How is the state here?

dragonwocky:
I've just pushed [a couple more features](https://github.com/dragonwocky/listmonk/blob/master/docs/docs/content/roles-and-permissions.md#list-roles) necessary for my use case. In summary:

- The `lists:create` permission gives users scoped ability to create lists for their own list role.
- Enabled messengers (e.g. SMTP servers) are configurable per-list role, with an override permission (`messengers:get_all`) and default access to the `email` messenger if all others are disabled.
- Default \`from\` addresses are configurable per-SMTP server.

@xeruf there are a couple of semi-significant changes here, so in terms of official release depends on @knadh's plans but hopefully 5.1.0? In the meantime I'm using this Dockerfile:

```
FROM golang:1.24-alpine AS builder
RUN apk update && apk add make nodejs yarn git
RUN mkdir -p /listmonk && cd /listmonk \
  && git clone --depth 1 https://github.com/dragonwocky/listmonk.git . \
  && make dist
FROM alpine:latest
WORKDIR /listmonk
COPY --from=builder /listmonk/listmonk .
COPY --from=builder /listmonk/config.toml.sample config.toml
COPY --from=builder /listmonk/docker-entrypoint.sh /usr/local/bin/
RUN chmod +x /usr/local/bin/docker-entrypoint.sh
ENTRYPOINT ["docker-entrypoint.sh"]
CMD ["./listmonk"]
```

The next thing might to be figure out some sort of scoping or per-user defaults for templates.

xeruf:
thanks, I think I'll be giving it a try in the course of this month

Right now I am still on v4.1.0 and one challenge I have is those same permissions for campaigns - i.e. I cannot give a user access to only own campaigns or of specific lists but maybe that's already solved with v5?

knadh:
Hi @dragonwocky. I'll review this sometime soon. Could you please:
- Fix the merge conflicts?
- Remove changes unrelated to the `list:create` permission? (Makefile, Dockerfile, messenger related changes, per-SMTP `from` e-mail). It would be best to consider the messenger changes and per-SMTP `from` as separate features/PRs.

dragonwocky:
@knadh I'll update this soon, sorry! Just finished semester so trying to catch up on a few things at the moment.

Issue #2295, opened 2025-02-07
Support for invisible hCaptcha


candidexmedia:
### **Is your feature request related to a problem? Please describe.**
I recently learned about invisible captchas, and found out that [hCaptcha supports them](https://docs.hcaptcha.com/invisible/). Implementing it on the embedded form on my website was fairly straightforward and works beautifully! The initial implementation of hCaptcha was clunky, and didn't mesh well with the website design. This option is a lot more seamless.

![Image](https://github.com/user-attachments/assets/8801eb33-3b71-4376-b40f-5af1048fa126)

### **Describe the solution you'd like**
I would love to see an option for invisible captchas in the provided Listmonk embed code (this could be an additional checkbox in the Settings > Security captcha section, or a checkbox in the Form page).

To implement it, I had to do the following (updated instructions):

1. Add an `id` attribute to the Listmonk form

2. Add an `id` attribute to the submit button

3. Add `data-callback="onSubmit"` and `data-size="invisible"` attributes to the hcaptcha `<div>`

4. Add an additional script before the hosted one

The final form code looks like this:

```html
<form method="post" action="https://mylistmonkinstance.com/subscription/form" target="_blank" class="listmonk-form" id="listmonk-form">
  <div>
        <h3>Subscribe</h3>
        <input type="hidden" name="nonce" />
    
        <p><input type="email" name="email" required placeholder="E-mail" /></p>
        <p><input type="text" name="name" placeholder="Name (optional)" /></p>
    
        <p>
          <input id="321" type="checkbox" name="l" checked value="54321" />
          <label for="321">My List</label>
        </p>

    <div class="h-captcha" data-sitekey="1234567" data-callback="onSubmit" data-size="invisible"></div>

    <input type="submit" value="Submit" id="listmonk-form-submit">
            
    <script type="text/javascript">
        const form = document.getElementById('listmonk-form');

	function validate(event) { // add back client-side validation
		event.preventDefault();
		if (form.checkValidity()) {
			hcaptcha.execute(); // if the form is valid, run captcha
		} else {
			form.reportValidity(); // if the form is invalid, display the error messages
		}
	}

	function onLoad() {
		var element = document.getElementById('listmonk-form-submit');
		element.onclick = validate; // run the validate function when the submit button is clicked
	}

	function onSubmit(token) {
		form.submit(); // submit the form if the hcaptcha is succesful
	}
    </script>

[...]

PR #2473, opened 2025-05-14
Support for Invisible hCAPTCHA

candidexmedia:
This PR adds support for [invisible hCAPTCHAs](https://docs.hcaptcha.com/invisible/) by adding a setting toggle that enables it.

Invisible hCAPTCHAs hide the CAPTCHA challenge until the user hits the submit button, and displays the CAPTCHA as a floating window. This allows the unstyled HTML form to blend in better into an existing website. (See https://github.com/knadh/listmonk/issues/2295)

I'm still new to this, so please let me know if a different approach should be taken.

Example demo of invisible CAPTCHA:

demo clip

### DIFF
--- a/frontend/src/views/Forms.vue
+++ b/frontend/src/views/Forms.vue
@@ -67,7 +67,7 @@ export default Vue.extend({
 
   methods: {
     renderHTML() {
-      let h = `<form method="post" action="${this.serverConfig.root_url}/subscription/form" class="listmonk-form">\n`
+      let h = `<form method="post" action="${this.serverConfig.root_url}/subscription/form" class="listmonk-form" id="listmonk-form">\n`
         + '  <div>\n'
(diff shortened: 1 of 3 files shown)


knadh:
Hi @candidexmedia. The PR seems to be broken with syntax errors. Also, I'm planning of integrating https://altcha.org and phasing out hCaptcha in the future.

candidexmedia:
> Also, I'm planning of integrating https://altcha.org/ and phasing out hCaptcha in the future.

I see that altcha has a [floating UI option](https://altcha.org/docs/v2/floating-ui/), so that's promising. How will it be integrated into Listmonk (since there's no hosted cloud version)?

knadh:
Closing this as I'm planning to integrate Altcha. It'll be self-contained within listmonk.

Issue #2541, opened 2025-06-27
List import count off by one.


howardah:
**Version:**
 - listmonk: [eg: v5.0.1]
 - OS: [Ubuntu Server]

**Description of the bug and steps to reproduce:**
When importing a csv list, it seems that the total count is one less than it should be. Which produces an inaccurate import screen. In the screenshot, for example, it shows 6/5 imported and 120%.

It looks like the problem comes from the `countLines` function in `internal/subimporter/importer.go` which produces a total count by counting the line breaks (`\n`) in the csv. This called in the `LoadCSV` function of the same file and one is removed from the total in order to exclude the header. This works assuming that the csv includes a trailing line break but doesn't whenever there is no trailing line break (as in the csv used in my example).

It's a small and purely aesthetic issue, but I would be happy to open a small PR to patch the `countLines` function to slightly more accurately count the lines by checking if there are trailing line breaks or not!

**Screenshots:**
<img width="888" alt="Image" src="https://github.com/user-attachments/assets/71c170e6-261c-420e-a924-7c7506088846" />

knadh:
Yep, makes sense. Thanks!

candidexmedia:
I'm having the same issue. My csv contained the `email, name` headers, and it seemed to have counted that as well?

I had 7 contacts to import, and it said there were 7/9 records, and 1 invalid email:

![Image](https://github.com/user-attachments/assets/1437e395-ac8d-4c8e-a79d-2d39e6e6e16b)

It would have been nice to also know which email(s) were invalid, and have them listed out to fix.

howardah:
@candidexmedia, the way that the rows are currently counted is just by counting all the line breaks (`\n`) and then subtracting 1 for the header row. I believe that this logic assumes a trailing line break at the end of the file and since my csv had no trailing line break, my 'total rows' count was one less than it should be.

Given that your row count is, as it sounds, one more than it should be, it seems possible that you have more than one trailing line break in your csv or possibly another line break in the middle of the file. Does that sound right?

The PR that I opened is really just a small change to add one to the count if there is no trailing slash. But maybe, it's worth re-addressing the logic of counting the rows so that empty rows aren't counted at all?

Providing feedback about rows that failed to process is a good idea, but seems out of the scope of this "issue", so maybe that's worth opening a new one? Does that seem fair @knadh ?

howardah:
[...]

PR #2542, opened 2025-06-27
Update countLines for improved accuracy

howardah:
Closes issue #2541 

Just a small change to the `countLines` function which just checks if the last byte is a line break (`\n`) and adds one to the total count if it's not. This means that the row count remains more accurate if there is no trailing line break.

This does not introduce any logic to distinguish between "blank" and non "blank" lines and therefore still does not guarantee an accurate row count (even just in the case of multiple trailing line breaks), but does approach a _more_ accurate row count.

### DIFF
--- a/internal/subimporter/importer.go
+++ b/internal/subimporter/importer.go
@@ -707,11 +707,10 @@ func (s *Session) mapCSVHeaders(csvHdrs []string, knownHdrs map[string]bool) map
 // Credit: https://stackoverflow.com/a/24563853
 func countLines(r io.Reader) (int, error) {
 	var (
-		buf       = make([]byte, 32*1024)
-		count     = 0
-		lineSep   = byte('\n')
-		lastByte  byte
-		bytesRead int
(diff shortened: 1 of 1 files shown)


internal/subimporter/importer.go
```diff
@@ -707,23 +707,34 @@ func (s *Session) mapCSVHeaders(csvHdrs []string, knownHdrs map[string]bool) map
...
 		c, err := r.Read(buf)
-		count += bytes.Count(buf[:c], lineSep)
-
-		switch {
-		case err == io.EOF:
-			return count, nil
+		if c > 0 {
+			count += bytes.Count(buf[:c], []byte{lineSep})
+			lastByte = buf[c-1]
+			bytesRead += c
```
knadh, line 722: Since `bytesRead` is only checked once (`if bytesRead > 0`)  as a boolean, it's not necessary to count the bytes here. For simplicity, won't simply replacing `if bytesRead > 0 ..` with `lastByte != 0` work?

Issue #250, opened 2021-01-20
Send campaign to segment of list


nickwynja:
This feature would allow using queries to send to a segment of a list. This is a pretty standard approach in email service providers. This is somewhat related to #129. The problem I see with creating dynamic "query lists" is that when someone wants to unsubscribe, they might only be unsubscribed from that list and not the "master" list.

To avoid this, I'd like to send a campaign to a segment. For example, I'll add a customer to the master list when they register and include a custom attribute with the subscribe date. Then I could send a campaign to a sub-segment who have registered in, for example, the last 30 days based on a SQL query. The campaign send would still be considered a part of the master list so that an unsubscribe would be unsubscribed from the master list.

Thanks for making listmonk! I've been looking for something like it for a long time.

knadh:
Thanks @nickwynja.

The problem with creating a sub list based on a query is that the results can include subscribers from multiple lists, in which case, it would not possible to have a single parent list.

nickwynja:
I see. I think this could be managed by the segment being a sub-query just like is happening on `/subscribers/lists/<id>` page now. When you query on that page, it appears to only be querying against the list. Here I can search for names or attributes and results are only showing members of the current list.

> Thanks @nickwynja.
> 
> The problem with creating a sub list based on a query is that the results can include subscribers from multiple lists, in which case, it would not possible to have a single parent list.

knadh:
Closing this as there's a longer discussion on the same matter here https://github.com/knadh/listmonk/issues/250.

nickwynja:
@knadh In your comment, you referred to this (self) issue. Did you mean to refer to a different ticket or did you mean to comment and close a different ticket?

alerque:
Please re-open this issue (or link to whatever issue is actually tracking this). It is still a needed feature and the linked discussion issue is ... self referential.

knadh:
@nickwynja @alerque looks like I've lost track of the issue I really meant to reference, and now I'm unable to find it either. Re-opening this.

alexbrazier:
Would it be possible to do something here to filter the list based on a query/segment https://github.com/knadh/listmonk/blob/3386de40c74740b2577e5344e9f6456e9ba50502/queries.sql#L635-L637

```sql
 WHERE subscriber_lists.status != 'unsubscribed' AND
    ## Either grab a query saves in a segments table or get camps.query directly
    ## Not sure if it's possible to convert the string into a real WHERE condition though
[...]

PR #2546, opened 2025-06-28
Add SQL snippets functionality (part 1 of dynamic subscriber lists)

restyler:
This is part 1 of implementing "segments" or "dynamic subscriber lists" (https://github.com/knadh/listmonk/issues/250). This PR adds the foundational SQL snippets functionality that is an isolated (and useful) feature itself and later this can be used by dynamic subscriber lists in a future PR.

The idea of "dynamic subscriber list" would be that it is just another "subscriber list" but it is dynamically updated via cron scheduler (and right before campaign is sent), by executing "sql snippet" attached to this list. I am pretty sure there will be some rough edges here, but at least this is something to start the discussion.

## Summary
- Adds SQL snippets database table and CRUD operations
- Provides Vue.js UI for managing reusable SQL query fragments
- Includes SQL validation against allowed tables
- Adds count functionality to test snippet queries

## Context
Splitting the dynamic segments feature into two parts:
1. **SQL snippets in database** (this PR) - foundational reusable query system
2. **Dynamically updated segments** (future PR) - automatic list membership management


restyler:
Video demo:

https://github.com/user-attachments/assets/c040d8b3-c4b2-4544-b52c-e0d1494df514

relikd:
Hahaha, nice to see my idea (https://github.com/knadh/listmonk/issues/250#issuecomment-1821915489) implemented :D thank you!

Regarding your video demo, it looks like the snippet is just pasted in the subscribers list. Maybe it would be better if the filter would use the snippet reference? E.g., if I update the snippet, all occurrences are updated as well. Or is it intentional? If so, what is your imagined scenario? An admin pre-configuring some sample queries for users to be able to edit them?

Thanks for the progress :)

restyler:
Thanks!

> E.g., if I update the snippet, all occurrences are updated as well. Or is it intentional? 

This is a great question. I want to minimize the stress for current users (which might be used to editing snippets right here, in textarea at "Advanced" section) and add "progressive enhacement" to the textarea instead of strong reference to SQL snippets.

I have just added a way to quickly save snippet from Subscribers page - and live counter.
2025-06-29 at 01 11

knadh:
Thanks for the PR @restyler! Looks fine prima-facie. I'll review it closer to the next big release.

About the live counter though, I'm not sure if it's ideal. Executing queries continuously on large instances (ours at work has ~17+ million subscribers) just to get the count is going to be very problematic, especially when it's as you type. Also, from a UX perspective, the count is just one-click away anyway.

restyler:
>  I'll review it closer to the next big release.

Oh, this would be so nice to have this included in next major release! Thanks!

 > Executing queries continuously on large instances (ours at work has ~17+ million subscribers) just to get the count is going to be very problematic, especially when it's as you type. 

I agree this might be a problem on huge lists. That said, I suspect such extreme cases are quite rare, and for them live editing can be easily disabled with a single checkbox (I think this setting state should be stored, so it is disabled once and forever), and I remember that when I started writing SQL queries in Listmonk, my biggest pain was the lack of interactivity and the lack of transparency around what “attribs” my subscribers actually have (this could be solved by some lightweight sniffing feature - again, it could be a bit slow on 2% of listmonk installations - but for another 98% this would mean much easier adoption and usage).

ilkkao:
Also there could be a "Execute snippet" button that updates the number. This would be similar e.g. how formula editing in Grafana dashboards work.

Rush:
I’m considering adopting Listmonk, and the addition of this feature (dynamic subscriber lists) would make it an easy decision to switch.

Example use case: localized newsletters and user selecting their language of choice in the application. 

With this feature, it'd be a simple case of querying for the right language property.

Without this feature, the application has to move the user between lists.

filipesmedeiros:
@knadh Without necessarily committing to it, I wonder if it is possible to sponsor specific features? Would you be open to crowdfunding this feature? I think there would be enough people pooling resources behind it. I can talk about it to my organization, maybe other people like @Rush would as well?

NWTrident:
Any updates on this?  Is there another PR for creating dynamic lists?

PR #2558, opened 2025-07-08
external HTML support

Gawkky:
This is a functionality which makes it automatically generated content to be included without use of iFrames.

This can be done with AWS Lambda + AWS Cloudfront (Both indefinitely free).

### DIFF
--- a/internal/manager/manager.go
+++ b/internal/manager/manager.go
@@ -4,7 +4,9 @@ import (
 	"errors"
 	"fmt"
 	"html/template"
+	"io"
 	"log"
+	"net/http"
 	"net/textproto"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


Gawkky:
So I've split the feature up in 2 parts:

1. ExternalHtmlDynamic
2. ExternalHtmlStatic

Dynamic renders for every subscriber the content again, this is useful when generating user specific content based on name, email, UUID or other specifics available in listmonk.

Static on the other hand renders the html once and keeps it stored in a cache. This is handy for unpersonalized but yet automatic generated content.

Both of the functions copy the source code and do not require the user to load in an external snippet stored on some CDN or something. This makes that keeps its content even when overriding the external content itself.

knadh:
Thanks for the PR @Gawkky, but this seems like a very niche feature that requires changes to the campaign manager. I don't think this is ideal to be merged into the core.

Gawkky:
Maybe we could put it in the settings and let the user itself decide if it wants to enable it?
I feel like this is gives the end user the freedom to build more complex and event driven newsletters.

My use case for example is that I want to show the latest three articles on our site without everytime doing it by hand. There is a lambda function that generates and creates the html snippet, which can be embedded.

I also see the possibility for some users to personalize standard campaigns as transactional mails don't really tick that box.

knadh:
hm, but this HTML-pull is a niche, unorthodox way of automating e-mail body creation. The recommended approach is to use the listmonk APIs to "push" the content. Your lambda can run, prepare the content, and simply curl the campaign into listmonk via its APIs.

There are also tools like this for example: https://github.com/ping13/listmonk-rss

PR #2594, opened 2025-07-25
feat: custom list unsubscribe

libo:


knadh:
Hi @libo. What does this PR do?

libo:
Hi! @knadh 
Sorry I never had the time to finish it yet.

The idea is to allow for a custom list unsubscribe. If you are for instance using an external CRM system to manage your list and subscription. This will allow to customize the URL for the unsubscribe link and to allow for expansion of attributes.

So if you import user with a non-monotonic and unguessable id that the CRM can use to identify the user you could have a unsubscribe url like:

```
https://acme.com/unsubscribe/user_id={{ .Subscriber.Attribs.custom_id }}

knadh:
Hi @libo. This is too niche a feature though, customizing unsubscription IDs. I don't think it makes sense in the core.

The simpler way here is to use the listmonk APIs and script the desired behaviour elsewhere (CRM or any other public facing page).

Thank you.

PR #2600, opened 2025-07-28
chore: add nginx service to docker-compose upgrade flow

ankitkhadria:
Docker upgrade flow to explicitly include the nginx service during restarts


knadh:
Hi @ankitkhadria. There is no `nginx` service in the official reference file: https://github.com/knadh/listmonk/blob/master/docker-compose.yml

Looks like this is specific to your setup.

PR #2614, opened 2025-08-05
Optimized Dockerfile and docker-compose.yml

yashsinghshiva:
Added a dedicated run.sh script for cleaner entrypoint handling. Also updated docker-compose.yml by adding a healthcheck for the database service and removing the unnecessary 'command' section from app definition.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,7 +6,7 @@ x-db-credentials: &db-credentials                             # Use the default 
 services:
   # listmonk app
   app:
-    image: yashsingh/listmonk-init:latest
+    image: listmonk/listmonk-init:latest
     container_name: listmonk_app
     restart: unless-stopped
     ports:


stcalica:
loving this cause the Dockerfile needs to be more "self-contained" and needs a better developer experience

docker-compose.yml
```diff
@@ -6,7 +6,7 @@ x-db-credentials: &db-credentials                             # Use the default
 services:
   # listmonk app
   app:
-    image: listmonk/listmonk:latest
+    image: listmonk/listmonk-init:latest
```
knadh, line 9: What is `listmonk-init`? That is a non-existent image.

Dockerfile
```diff
@@ -3,24 +3,21 @@ FROM alpine:latest
...
-
-# Copy the entrypoint script
+# Copy binaries and configs
+COPY listmonk run.sh config.toml config.toml.sample queries.sql schema.sql permissions.json ./
 COPY docker-entrypoint.sh /usr/local/bin/
 
-# Make the entrypoint script executable
-RUN chmod +x /usr/local/bin/docker-entrypoint.sh
+# Copy static assets
+COPY static/ ./static/
```
knadh, line 14: This is incorrect. These assets are bundled into the binary during the build step and have no place in the Docker image.

run.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+# Run database installation only once
+./listmonk --install --idempotent --yes
```
knadh, line 4: This is incorrect. `--config=''`, `--upgrade=''` etc. are necessary for listmonk to auto-updated on boot when the image version changes. Why were they removed?

knadh:
Hi @yashsinghshiva. This PR has several breaking issues. Just wondering, was it auto-generated by an LLM? I'm afraid I've to close this as I don't have the bandwidth to review further as the proposed changes don't bring any quantifiable benefits. Thank you.

yashsinghshiva:
Hey @knadh ,  The reason for this PR is that I was attempting to self-host Listmonk on a PaaS platform, where I ran into limitations in passing the complete startup command:
command: [sh, -c, "./listmonk --install --idempotent --yes --config '' && ./listmonk --upgrade --yes --config '' && ./listmonk --config ''"]
To work around this, I created a run.sh script and made corresponding adjustments in docker-compose.yml. I also built a local Docker image (listmonk-init) to test and deploy.

While I did use an LLM to help me understand parts of the codebase, the changes themselves were made specifically to adapt Listmonk for my deployment environment’s constraints.

I’ll review this in detail, make the necessary changes, and submit a new PR accordingly.

knadh:
Ah, it was a custom setup for your environment. Since any blockers pertaining to the `command` in the listmonk setup hasn't cropped up elsewhere, I don't think any changes are necessary right now. Thank.

PR #2643, opened 2025-08-31
WIP: Add MJML support.

knadh:


Rush:
Nice, it'd be pretty great to have! What's remaining to merge it?

knadh:
The MJML lib version that I tested this branch was broken, but it's improved since then. That dependency has to be updated and the implementation has to be tested fully. If you're able to do it, that would be great!

knadh:
I've updated the `mjml` branch with the latest version of the mjml lib. Test Linux build is here: [listmonk.tar.gz](https://github.com/user-attachments/files/25787487/listmonk.tar.gz)

Someone has to test this out and help before I can merge it :)

preslavrachev:
@knadh feel free to bump up to v0.11.0 : https://github.com/preslavrachev/gomjml/releases/tag/v0.11.0

preslavrachev:
@knadh I forked the entire repo, and will issue a separate PR to merge my `mjml` changes into yours. There is a bit of a chicken-and-egg problem I am trying to hopefully solve soon. In short, listmonk, assumes that people create campaigns which may reuse the same template. The issue is that MJML compilers cannot compile and validate partial MJML code, you need to combine everything - the campaign, wrapped inside the template, before you pass it onto the MJML compiler. It's the opposite of what's currently happening in listmonk, and I am working on an adaptation strategy.

knadh:
@preslavrachev I've addressed this and force pushed to the `mjml` branch. Please give it a shot.
- The base template can now have `<mjml> ... {{ template "content" . }} </mjml>`
- Or it can just have `{{ template "content" . }}` and expect the full mjml body in the campaign.

dfabulich:
I've pulled down the `mjml` branch and I have a bug to report. (I could file a PR on this, too, if you'd like.)

There's an error message if you select "MJML" on the first page of the Campaign wizard. "Error compiling campaign body: error compiling MJML: failed to parse MJML: EOF".

At that point, the campaign body is blank (of course). It should tolerate an empty body until I get to the second tab.

knadh:
@dfabulich amended the branch with a fix. Verified campaign creation and type conversion.

dfabulich:
In https://github.com/knadh/listmonk/pull/3131 I've fixed the first P1 issue Hodor raised.

Hodor's second P1 issue is quite confusing, but I don't think it's a real issue.

> **[P1] Fix content conversion paths involving MJML (returns non-MJML / can't handle partials)** (`models/campaigns.go:272-297`)
> * `ConvertContent()` treats `markdown -> mjml` as `markdown -> HTML` (returning plain HTML but labeling it MJML), which will typically fail later when compiling/saving a MJML campaign because the result isn't a valid MJML document/fragment. Additionally `mjml -> html|richtext` renders only `c.Body`, so it will error for the common case where the campaign body is a _partial_ MJML fragment intended to be wrapped by a selected base MJML template (the frontend format-switch uses this endpoint).

This is actually two issues:

1. `ConvertContent()` treats `markdown -> mjml` as `markdown -> HTML` (returning plain HTML but labeling it MJML)
2. `mjml -> html|richtext` renders only `c.Body`, so it will error for the common case where the campaign body is a _partial_ MJML fragment intended to be wrapped by a selected base MJML template (the frontend format-switch uses this endpoint).

I think the second issue is simply incorrect; I tested partial MJML with the provided "Sample MJML template" and it worked just fine, setting the campaign MJML to `<mj-text>hello world</mj-text>`. The campaign sent, and, when converting from MJML to HTML or to Richtext, it left the `<mj-text>` element untransformed, which seems fine to me; HTML just treats unknown elements as `<span>s`. (The user is warned that the transformation is lossy.)

The gotcha is that embedding HTML or partial MJML inside an MJML template requires you to know/manage the current context element. The provided "Sample MJML template" embeds the email content in an `<mj-column>` inside an `<mj-section>` inside `<mj-body>`, like this:

```xml
<mjml>
  <mj-body background-color="#F0F1F3">
    <!-- Main Content Wrapper -->
    <mj-section background-color="#fff" border-radius="5px" padding="30px">
      <mj-column>
        {{ template "content" . }}
      </mj-column>
    </mj-section>
  </mj-body>
</mjml>
```

If your partial MJML is just text like `Hello world`, that won't render properly. You're not allowed to just drop text in an `<mj-column>`; you need to put your text in an `<mj-text>` element.

But Listmonk has no way of knowing how users will construct their templates. If, instead, the template had included an `<mj-text>` element, like this, `Hello world` would render fine:

```xml
<mjml>
  <mj-body background-color="#F0F1F3">
    <!-- Main Content Wrapper -->
    <mj-section background-color="#fff" border-radius="5px" padding="30px">
      <mj-column>
        <mj-text>
          {{ template "content" . }}
        </mj-text>
      </mj-column>
    </mj-section>
  </mj-body>
</mjml>
```

Arbitrary HTML renders fine inside an `<mj-text>`. So, if the template provided its own `<mj-text>` element, typing stuff in the Rich Text editor or the Markdown editor and "Converting to MJML" by just rendering it as HTML would work fine; the HTML -> MJML "converter" would be a no-op.

But a template like that would also be very confining. If you wanted to add more sections/columns to your MJML campaign, you'd first have to close out the tags that the template provided, and then reopen tags for the template to close, like this:

```
</mj-text>
</mj-column>
</mj-section>

<!-- now include your custom section -->
<mj-section>
  <mj-column>
    <mj-text>hello</mj-text>
  </mj-column>
</mj-section>

<!-- now you have to reopen your tags again so the template can close them -->
<mj-section>
<mj-column>
<mj-text>
```

Overall, it's not at all clear to me what a Listmonk "Sample MJML Template" should provide, but I suspect that it should either provide its own `<mj-text>`, which means converting from HTML to MJML would be a no-op, or it should only provide an `<mj-body>`, with an `<mj-head>` that sets `<mj-attributes>` to provide default attributes for `<mj-section>`s.

If the template only provides an `<mj-body>`, then, when converting from Markdown / HTML to MJML, _somebody_ would need to wrap the resulting MJML in `<mj-section><mj-column><mj-text>` tags so it will render. That could be a manual step for campaign authors, or it could be in code, so the Markdown/HTML -> MJML converter could automatically add those tags.

Regardless, I don't think this is a P1 issue. With my PR #3131 merged, I think MJML support is ready to ship.

PR #2802, opened 2025-12-05
Respect SOURCE_DATE_EPOCH when building

toabctl:
To make the build reproducable (see [0]), respect the SOURCE_DATE_EPOCH variable and don't include a build timestammp in the binary when set.

[0] https://reproducible-builds.org/


toabctl:
tested with:
```
❯ make dist     
CGO_ENABLED=0 go build -o listmonk -ldflags="-s -w -X 'main.buildString=v5.1.0 (#f8dee184 2025-12-05T07:48:45+0000)' -X 'main.versionString=v5.1.0'" cmd/*.go
```

vs.
```
❯ SOURCE_DATE_EPOCH=1 make dist 
CGO_ENABLED=0 go build -o listmonk -ldflags="-s -w -X 'main.buildString=v5.1.0 (#f8dee184)' -X 'main.versionString=v5.1.0'" cmd/*.go
```

knadh:
Hi @toabctl. If `SOURCE_DATE_EPOCH` is set, it should be used as the Unix timestamp offset and the `date` should be derived from that, otherwise, the system time. Here, if it's set, the date bit in the build string is completely eliminated, which isn't right.

knadh:
@toabctl would you be able to review this?

toabctl:
> @toabctl would you be able to review this?

I'll do during the next days. sorry for the delay

PR #2842, opened 2025-12-30
Convert Dockerfile to multi-stage build

dlackty:
This PR converts the Dockerfile to use a multi-stage build, enabling fully self-contained Docker builds without requiring local Go installation.

## Changes
- Add build stage using `golang:alpine` to compile the binary and frontend
- Install `stuffbin` and run `make dist` in the build stage
- Copy only the compiled binary to the runtime stage
- Remove `.gitignore` from `.dockerignore` (required for frontend eslint)

## Benefits
- Users can build the Docker image without installing Go locally
- Smaller final image size (excludes build tools and dependencies)
- Reproducible builds in CI/CD environments
- Follows Docker best practices for multi-stage builds

## Testing
Tested successfully with:
```bash
docker build -t listmonk:latest .
```

The build completes successfully and produces a working image.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,7 +2,6 @@
 **/.dockerignore
 **/.env
 **/.git
-**/.gitignore
 **/.project
 **/.settings
 **/.toolstarget
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,14 +1,31 @@
+# Build stage
+FROM golang:alpine AS builder
+
+# Install build dependencies
+RUN apk --no-cache add git make nodejs npm yarn
+
+# Set the working directory
+WORKDIR /app
+
+# Copy source code
+COPY . .
+
+# Build the application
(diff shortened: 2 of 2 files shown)


knadh:
Thanks @dlackty. Could you please test and confirm that this works with goreleaser? https://github.com/knadh/listmonk/blob/master/.goreleaser.yml

dlackty:
Thanks for the feedback @knadh. After reviewing the goreleaser configuration, I realize this change would break the existing release workflow since goreleaser expects the Dockerfile to work with pre-built binaries.

While a multi-stage Dockerfile would be useful for local development, it would require maintaining two separate Dockerfiles (one for goreleaser, one for dev), which adds complexity without significant benefit since the project already has a well-documented build process with `make dist`.

I'm closing this PR as the current approach is more appropriate for the project's release workflow. Thanks for taking the time to review!

PR #2863, opened 2026-01-13
Add setting to serve media from private s3

### LINKED ISSUE #2611: Serve media from internal/private S3 bucket via proxy
**Is your feature request related to a problem? Please describe.**
---
In our organization, we host a private MinIO instance for security reason. This instance is not publicly accessible. When we configure listmonk to use this MinIO for media, the generated links point to the MinIO instance's internal address, making the media inaccessible to our email recipients. While listmonk can be configured with a public bucket or CDN, this forces us to either expose our storage publicly or manage an additional service to handle a public-facing proxy, which adds unnecessary complexity.

---
Ideally, I would like to add an option to proxy the GET request for media to S3. Other services, such as GitLab, have this feature, meaning we wouldn't need an additional service to handle this requirement. Perhaps we could add a handler here to reuse the URL for a static file when the file provider is `s3`.

https://github.com/knadh/listmonk/blob/ba24c64fae557082c50798dacb5d1ad6b89a1ad5/cmd/init.go#L884-L887

While workarounds like setting up a separate reverse proxy are possible, they introduce additional maintenance overhead and complexity. A built-in feature would be more robust, easier to configure, and would provide a seamless, integrated experience for users with private S3-compatible storage.

Bogay:
This PR adds support for serving S3-uploaded media files directly through the application. The backend server will check whether upload URI is set for S3 media provider first to determine how to generate URL for media.

Fix #2611


knadh:
Thanks for the PR @Bogay.

I'm unable to understand `upload.s3.upload_uri` is for. Can't S3 uploads work as-is? Perhaps what's required is a config option that says ~ "Proxy media via listmonk URL"? If that's checked, the requests are proxied.

Bogay:
> Thanks for the PR @Bogay.
> 
> I'm unable to understand `upload.s3.upload_uri` is for. Can't S3 uploads work as-is? Perhaps what's required is a config option that says ~ "Proxy media via listmonk URL"? If that's checked, the requests are proxied.

Thanks for the quick review!

The `upload.s3.upload_uri` is used for configuring the path used for proxying requests to S3, just like how `upload.filesystem.upload_uri` works. Reusing the `upload.s3.public_url` might be acceptable, its validation rules would need to be updated. Now listmonk requires a full URL including domain and schema (`http://` or `https://`). But in the scenario of proxying media requests to S3, I think only paths make sense here (e.g. `/uploads`), or we need to extract the path part from `upload.s3.public_url`.

If we go with your suggestion of adding a config option for proxying, how should we handle the URL format? Do you prefer keeping a separate URI field, or updating the `upload.s3.public_url` validation to allow both paths and full URLs?

knadh:
> If we go with your suggestion of adding a config option for proxying, how should we handle the URL format? Do you prefer keeping a separate URI field, or updating the upload.s3.public_url validation to allow both paths and full URLs?

I think this is a better approach, as having both public URL + public upload URI would be redundant. We can have the field be either a `/path` starting with `/` or a fully formed URL. If it's a path, we implicitly register that handler and start proxying the requests. This can be explained in the help text below the field.

We will have to change the field's caption from `Custom public URL (optional)` to `Public URI or URL (optional)`.

knadh:
Hi @Bogay, if you can make that change, I can review and merge. Thank you.

Bogay:
Hi @knadh, sorry for late. I've finished the changes as per our discussion. Now we can configure S3 public URL to a path starts with `/` to make it proxying media requests to upstream S3. About the label of `upload.s3.public_url`, I settled on `Custom public URL or path (optional)` because I found the term URI seems to not precisely describe the value can be a path. [^1]

[^1]: https://developer.mozilla.org/en-US/docs/Web/URI

BTW, I found that currently the public URL setting is only validated at the frontend. Should I add validation logic to backend? Maybe another PR.

Issue #2867, opened 2026-01-15
json: cannot unmarshal string into Go struct field forwardemailNotif.truth_source of type bool


CodeAlDente:
**Version:**
 - listmonk: v6.0.0
 - OS: Linux ❤️ 

**Description of the bug and steps to reproduce:**

**Error message:**
```
bounce.go:214: error processing forwardemail notification: 
error unmarshalling Forwardemail notification: 
json: cannot unmarshal string into Go struct field forwardemailNotif.truth_source of type bool
```

Listmonk throws an error when processing bounce and complaint webhooks from ForwardEmail because it expects the `truth_source` field to be a `boolean`, but ForwardEmail sends it as a `string`, cf. https://forwardemail.net/en/faq#do-you-support-bounce-webhooks

**It is caused here:**
https://github.com/knadh/listmonk/blob/cd1bb1bc2045e8fb0827530947441e0721722d22/internal/bounce/webhooks/forwardemail.go#L34

knadh:
Ah. @shaunwarman this seems to have happened in PR #2250. Could you please take a look?

https://github.com/knadh/listmonk/pull/2250/files#diff-95f741a00f9b2f60e59e99d0acb8068681b848d88a8422693c84b384f490c88fL33

shaunwarman:
thanks @knadh @CodeAlDente - will take a look

PR #2869, opened 2026-01-16
fix: adjusted ForwardEmail webhook parsing to accept truth_source as e…

shaunwarman:
…ither a JSON boolean or string

Closes https://github.com/knadh/listmonk/issues/2867

### DIFF
--- a/internal/bounce/webhooks/forwardemail.go
+++ b/internal/bounce/webhooks/forwardemail.go
@@ -31,7 +31,7 @@ type forwardemailNotif struct {
 	Message         string            `json:"message"`
 	Response        string            `json:"response"`
 	ResponseCode    int               `json:"response_code"`
-	TruthSource     bool              `json:"truth_source"`
+	TruthSource     string            `json:"truth_source"`
 	Headers         map[string]string `json:"headers"`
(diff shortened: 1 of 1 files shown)


internal/bounce/webhooks/forwardemail.go
```diff
@@ -31,12 +32,34 @@ type forwardemailNotif struct {
...
-	TruthSource     bool              `json:"truth_source"`
+	TruthSource     boolString        `json:"truth_source"`
 	Headers         map[string]string `json:"headers"`
 	Bounce          BounceDetails     `json:"bounce"`
 	BouncedAt       time.Time         `json:"bounced_at"`
 }
 
+type boolString bool
+
+func (b *boolString) UnmarshalJSON(data []byte) error {
```
knadh: hm, this implementation seems like an overkill.

What is this field supposed to be? I saw in the docs that it's values like "yahoo.com". Does this mean that it can be "true", "false", or "yahoo.com" (all strings)? If yes, then we can just leave the field as `string` since it's not used in any logic anywhere, right?

shaunwarman: yep good call, will keep it simple

PR #2888, opened 2026-01-26
feat: Enhance link click registration to return campaign name

ilhamsj:
- Updated the `RegisterCampaignLinkClick` function to return both the destination URL and the associated campaign name in a single query.
- Modified the `LinkRedirect` method to handle the new return values, ensuring proper campaign name assignment.
- Adjusted the SQL query to fetch the campaign name alongside the URL, improving efficiency by reducing the number of database calls.


knadh:
Sorry, this is a big PR that changes too many things and I can't figure out what's going on or what it's supposed to do.

Please see: https://github.com/knadh/listmonk/blob/master/CONTRIBUTING.md#2-pull-requests

PR #2908, opened 2026-02-16
fix: handle NULL list_id in campaign processing to prevent crash

### LINKED ISSUE #2851: "list_id": converting NULL to int is unsupported
**Version:**
 - listmonk: [eg: v1.0.0]
 - OS: Docker Image

Starting a campaign resulted in an error:

```
2026/01/06 20:41:45.448222 manager.go:281: error processing campaign batch (Test campaign): error fetching campaign subscribers (Test campaign): sql: Scan error on column index 4, name "list_id": converting NULL to int is unsupported
```

and the campaign is still running (the list had only a single user)

**What I did:**
I deleted all the preexisting users and lists, I created two new list and I added one user which is part of one list, I created a new template and I had a Campaign Attribute. I starte the campaign for one specific list.

**Screenshots:**

SriHarsha-23:
**Description:**
Fixes a crash that occurs when NextSubscribers encounters a NULL list_id in the runningCamp struct.

The ListID field was previously defined as int, which caused a sql: Scan error when the database returned a NULL value (e.g. from deleted lists or inconsistent states).

**Changes:**

Updated runningCamp struct in cmd/manager_store.go: Changed ListID type from int to sql.NullInt64.

Updated NextSubscribers: Added a check for c.ListID.Valid before appending to the list of IDs.

**Related Issue:**
Closes #2851

### DIFF
--- a/cmd/manager_store.go
+++ b/cmd/manager_store.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"database/sql"
 	"github.com/gofrs/uuid/v5"
 	"github.com/knadh/listmonk/internal/core"
 	"github.com/knadh/listmonk/internal/manager"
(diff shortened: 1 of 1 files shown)


knadh:
Thanks for reporting the issue. The actual fix is much simpler and cleaner: https://github.com/knadh/listmonk/commit/0ee89f9097a0007ceef9aaf9e144b3845868fc8d

PR #2913, opened 2026-02-20
Add global setting to disable view and click tracking (#2912)

### LINKED ISSUE #2912: Feature Request: Global setting to completely disable view and click tracking
### Problem

Currently, listmonk offers two privacy-related mechanisms:

1. **Individual subscriber tracking toggle** (Settings → Privacy): anonymizes tracking by stripping subscriber UUIDs from tracking events, but still records aggregate view and click data.
2. **Template-level control**: admins can remove `{{ TrackView }}` and avoid `{{ TrackLink }}` in templates to prevent tracking pixels and link rewriting.

However, there is **no global admin setting to completely disable all tracking**. The only way to ensure no tracking occurs is to manually audit every template and remove the relevant tags. This is fragile and error-prone: any user who creates or edits a template can inadvertently re-enable tracking by adding `{{ TrackView }}` or `{{ TrackLink }}`.

### Use case

We use listmonk at a university to send email campaigns (e.g. course announcements, event invitations) to students and staff. Under GDPR, embedding tracking pixels in emails sent to university members without explicit consent is problematic. Even when the tracking is anonymized, it still constitutes processing of personal data (IP addresses, user agents, and behavioral data like open times are logged in HTTP requests).

We need a way to **guarantee at the infrastructure level** that no tracking occurs, regardless of what individual templates contain. Relying on template discipline across multiple campaign creators is not sufficient for regulatory compliance.

### Proposed solution

Add a global setting (e.g. in Settings → Privacy) such as **"Disable all tracking"** that, when enabled:

- Makes `{{ TrackView }}` render as an empty string (no tracking pixel is inserted into emails)
- Makes `{{ TrackLink "https://example.com" }}` pass through the original URL unmodified (no redirect/tracking URL is generated)
- Optionally stops recording view and click events on the backend, even if tracking endpoints are hit

This approach has several advantages:

- **Non-breaking**: Existing templates continue to work — they just silently stop tracking. No need to edit templates.
- **Centralized control**: A single admin toggle enforces the policy across all campaigns and templates.
- **Reversible**: If the organization later decides to enable tracking (with proper consent mechanisms), they can simply toggle the setting off.
- **Defense in depth**: Even if a template author adds `{{ TrackView }}` or `{{ TrackLink }}`, the system respects the global policy.

### Related issues

- #373: A user asked about completely disabling tracking; the response was to remove `{{ TrackView }}` from templates. No global setting was proposed.
- #193: Introduced the anonymous tracking toggle, which anonymizes but does not disable tracking.
- #1856: Requests splitting view and click tracking into separate toggles, which is complementary to this proposal.

### Additional context

This would make listmonk significantly more attractive for use in privacy-sensitive environments such as universities, public institutions, healthcare organizations, and NGOs, where GDPR or similar regulations require strong guarantees that no user tracking occurs unless explicit consent has been obtained.

krusche:
## Add global setting to disable view and click tracking

Closes #2912

### Summary

Adds a new boolean setting `privacy.disable_tracking` (default: `false`) under **Settings → Privacy** that completely disables email view and click tracking at the infrastructure level. When enabled:

- `{{ TrackView }}` renders as an **empty string** — no tracking pixel is inserted into emails
- `{{ TrackLink "url" }}` returns the **original URL unchanged** — no redirect/tracking URL is generated
- The backend tracking endpoints (`/link/...` and `/campaign/.../px.png`) **stop recording** views and clicks, while still redirecting old links so previously sent emails continue to work

This gives administrators a single, centralized kill-switch that enforces a no-tracking policy across all campaigns and templates, regardless of what individual template authors include. It is designed for privacy-sensitive environments (universities, public institutions, healthcare) where GDPR or similar regulations require infrastructure-level guarantees that no user tracking occurs.

#### Server side

| File | Change |
|------|--------|
| `models/settings.go` | Added `PrivacyDisableTracking` field to the `Settings` struct |
| `schema.sql` | Added `privacy.disable_tracking` default for new installs |
| `internal/migrations/v6.1.0.go` | New migration that inserts the default setting for existing installs (`ON CONFLICT DO NOTHING`) |
| `cmd/upgrade.go` | Registered `v6.1.0` migration |
| `cmd/init.go` | Added `DisableTracking` to `Config.Privacy` struct and passed it to the campaign manager config |
| `internal/manager/manager.go` | Added `DisableTracking` to manager `Config`; `TrackView` returns empty and `TrackLink` returns original URL when enabled; defensive guard added in `trackLink()` method |
| `queries/links.sql` | Added `get-link-url` query to resolve a link's URL without recording a click |
| `models/queries.go` | Added `GetLinkURL` prepared statement |
| `internal/core/campaigns.go` | Added `GetLinkURL()` method for click-free URL resolution |
| `cmd/public.go` | `RegisterCampaignView` returns the pixel without recording when tracking is disabled; `LinkRedirect` resolves and redirects the URL without recording a click |

#### User Interface

| File | Change |
|------|--------|
| `frontend/src/views/settings/privacy.vue` | Added "Disable e-mail tracking" toggle above the existing individual tracking toggle; individual tracking switch is disabled (greyed out) when the global toggle is on, since it becomes redundant |
| `frontend/src/views/CampaignAnalytics.vue` | Shows an informational message ("E-mail tracking is globally disabled") on the analytics page when the setting is active |
| `i18n/en.json` | Added translation strings for the new toggle, its help text, the disabled-state hint, and the analytics banner |

### Design decisions

- **Follows existing patterns**: The implementation mirrors exactly how `privacy.individual_tracking` is threaded through the codebase — same Config struct placement, same template function pattern, same migration pattern.
- **Non-breaking**: Existing templates continue to work unchanged. The tags simply become no-ops. No template edits are required.
- **Backwards compatible**: Old emails that were sent with tracking URLs before the setting was enabled still redirect correctly — the `/link/...` endpoint resolves the URL and redirects, it just skips recording the click.
- **Defense in depth**: Tracking is blocked at three layers: (1) template rendering (no tracking URLs/pixels generated), (2) the `trackLink` method (defensive guard), and (3) the HTTP endpoints (no DB writes even if old tracking URLs are hit).
- **Clear UX**: When the global toggle is on, the individual tracking toggle is disabled with an explanatory message. The analytics page shows a banner so admins understand why data may be empty.

### Test plan

- [ ] `make build` compiles without errors
- [ ] `go test ./...` passes
- [ ] `go vet ./...` passes
- [ ] Start the app, go to **Settings → Privacy** — verify the new "Disable e-mail tracking" toggle appears above individual tracking
- [ ] Enable the toggle, save — confirm the app restarts and the setting persists
- [ ] Verify the "Individual subscriber tracking" toggle is greyed out with an explanatory message
- [ ] Create a test campaign with `{{ TrackView }}` and `{{ TrackLink "https://example.com" }}` in the body
- [ ] Preview/send the campaign — verify no `` tracking pixel is inserted and links point directly to `https://example.com`
- [ ] Visit **Campaign Analytics** — verify the "tracking is globally disabled" message appears
- [ ] Disable the setting — verify tracking works normally again (pixel inserted, links wrapped)
- [ ] Test that old tracking URLs from previously sent emails still redirect correctly (the `/link/...` endpoint returns a redirect without recording)
- [ ] Run `--upgrade` on an existing database — verify the `v6.1.0` migration runs and adds the setting

### Screenshots

image


krusche:
Tested locally with the Docker dev environment (PostgreSQL + Mailhog).                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                             
With `privacy.disable_tracking` enabled:                                          
- `{{ TrackView }}` renders as empty: no tracking pixel in the email
- `https://listmonk.app@TrackLink` renders as plain `https://listmonk.app`: no redirect wrapping
- Manually added links (e.g. `https://www.google.de/@TrackLink`) also render as plain URLs
- The "Individual subscriber tracking" toggle is greyed out with an explanatory message

With `privacy.disable_tracking` disabled:
- Tracking pixel (px.png) is present in the email body
- Links are wrapped with /link/... tracking redirects
- Individual tracking toggle works independently as before

Both states verified by inspecting raw email source in Mailhog.

knadh:
Thanks for the PR, @krusche. I made a few changes including a bug fix, but was unable to amend the PR here as you sent it from your `master` branch. Copied and merged here instead: https://github.com/knadh/listmonk/pull/2920/commits

PR #2919, opened 2026-02-22
Add partial Croatian language localization

kreshot:
Focused on subscriber-facing messages


knadh:
Hi @kreshot. Do you think you would get a chance to finish this PR? Thank you.

PR #2924, opened 2026-02-23
Fix right-aligned 'Link Clicks' heading on dashboard.

shubhamatkal:
Removes the `has-text-right` class from the second chart heading so it aligns consistently with the 'Campaign Views' heading on the left.


knadh:
It's a deliberate design choice :)

Issue #2893, opened 2026-02-01
Allow changing email addres via `/subscription`


WilliamDEdwards:
Some users might want to change their email address, rather than having to unsubscribe and re-subscribe.

Via the `/subscription` route, one can change their name, but not email address.

bentuna:
This would be a nice feature and save us quite some work. A lot of subscribers write us and we have to change their address manually.

pranavdeepak13:
Hey Team, 
Can you assign me this issue? It seems like a good first issue to be contributing to. 
Looking forward to resolving this issue soon.

Regrads,
Pranav Deepak

dr-msr:
And may also update their name, too. 

Just checking if @pranavdeepak13 has made any progress, maybe I can chip in too

knadh:
Name change is already possible under "Manage preferences" on the unsub page. This can be enabled/disabled in Admin -> Settings

dr-msr:
Ah, awesome. I just realized the API allows updating all fields, including the attribs. This is very handy.

bentuna:
Would still be worth keeping this open.

PR #2933, opened 2026-03-03
Minimal implementation of a PATCH endpoint for subscribers.

zkirill:
This minimal implementation focused on a common scenario where a subscriber updates their email address in an external system and that change must be reflected in ListMonk.

I think this endpoint is sorely needed as a safer alternative to the existing `PUT` endpoint which has the ability to destroy `List` associations.

Please see related issue https://github.com/knadh/listmonk/issues/2893.


knadh:
Done here: https://github.com/knadh/listmonk/commit/24817d5a9384be47529597ed877f7700a0832a63

PR #2953, opened 2026-03-13
Add per-domain SMTP routing

jaymzh:
Currently, you can add multiple SMTP servers, but they will be
round-robin'd.

It is useful to be able to pick an SMTP server based on the `from`
of the campaign.

For example, Mailgun will DKIM sign based on the domain of the
authenticating user, so if you send campains from more than one domain,
you need to have different SMTP servers setup for each domain.

This adds an optional list of domains that each SMTP server should be
used with.

If multiple servers specify the domain in question, we will pick one
at random, just as the logic today does.

If no domains are specified, or non match, then we fall back
to the existing behavior of picking one at random.

Signed-off-by: Phil Dibowitz <phil@ipom.com>


jaymzh:
@knadh  - any thoughts on this?

knadh:
Haven't had a chance to review this yet. Will review in the coming couple of weeks. Thanks @jaymzh.

knadh:
Hi @jaymzh. I've refactored and cleaned up the PR a bit. Please give it a spin and once you confirm, we can merge.

jaymzh:
FYI: I had to do two things to make the dev containers work properly:
1. I had to modify `dev/app.Dockerfile` to based the image on node 22 instead of 16
2. I had to modify `frontend/vite.config.js` and add `host: '0.0.0.0',` to the `server` clause.

That said I was able to setup a somewhat contrived test instance where I created two smtp servers, with two from addresses, create two campains and try to send test messages from each and they used the correct smtp servers.

jaymzh:
Anything else needed on this @knadh ?

PR #2959, opened 2026-03-18
Fix POP bounce parsing failing on emails with UTF-8 BOM

aschwehm:
Some mail servers and proxies (notably Office 365 via [email-oauth2-proxy](https://github.com/simonrob/email-oauth2-proxy)) prepend a UTF-8 BOM (0xEF 0xBB 0xBF) to email messages. This causes the mail parser to fail with: "malformed MIME header key: ﻿Received"

The BOM bytes are interpreted as part of the first header name and making it invalid.

This fix strips the UTF-8 BOM from the raw message bytes before parsing. The bytes.TrimPrefix function is a no-op if the BOM is not present, so this change should be safe for all mail providers.

### DIFF
--- a/internal/bounce/mailbox/pop.go
+++ b/internal/bounce/mailbox/pop.go
@@ -1,6 +1,7 @@
 package mailbox
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


knadh:
Thanks @aschwehm. I think this is better fixed at the root, the POP3 lib that listmonk uses: https://github.com/knadh/go-pop3

PR #2961, opened 2026-03-20
Change wrapper min-height to 100%

anantanandgupta:
there was an annoying scrolling on the pages where the content was not long enough to cause the scroll.


knadh:
This breaks the layout and prevents the sidebar and the main content area from stretching to the end of the viewport.

Issue #2500, opened 2025-05-26
Max messages and duration at the SMTP level


mevinbabuc:
**Is your feature request related to a problem? Please describe.**
Lets say you have SMTP from AWS SES and Gmail configured. According to Gmail, the max emails you can send a day is 2000. where as, if its configured as SES the limit is very high.

**Describe the solution you'd like**
A solution would be to have max messages limit at the SMTP level

<img width="674" alt="Image" src="https://github.com/user-attachments/assets/77599d9c-422d-417a-9dba-cd60ce44e8e5" />

mevinbabuc:
True, that is what i'm doing currently, when i have different campaigns who use different SMTPs, i change the global limit, pause the other ones, reload. I keep doing this song and dance multiple times a day. A SMTP level config would be awesome. Global level can be like a default or overaching one.

Issue #2273, opened 2025-01-28
Feature: Support Amazon SES' throttled SMTP gateway by default


cmcaine:
**Is your feature request related to a problem? Please describe.**

tl;dr: an email campaign I sent was not delivered to around half the contacts because listmonk didn't understand amazon SES' throttling messages.

Longer: It's difficult to send mail from Amazon SES over SMTP because Amazon throttles requests by emails per second and listmonk:

1. does not allow us to throttle an SMTP connection in those terms
2. will try to send messages faster than a new SES account is likely to permit (with the default listmonk config)

If sending is throttled then some messages will not be sent on the campaign, even though they could be if listmonk would wait a little longer.

Amazon responds with a transient error when throttling: 454

> With the [SMTP](http://docs.aws.amazon.com/ses/latest/DeveloperGuide/SMTP.html) interface, Amazon SES responds with “454 Throttling failure: Maximum sending rate exceeded” after the DATA command.

[source](https://aws.amazon.com/blogs/messaging-and-targeting/how-to-handle-a-throttling-maximum-sending-rate-exceeded-error/)

**Describe the solution you'd like**

The best solution for me would be to make it real easy to run listmonk-messenger with listmonk, that way I would get complaints, etc coming back into listmonk.

A different good solution would be for listmonk to understand that a 454 error with the words "Throttling failure" means that a message should be re-queued without consuming a "retry" and maybe the sending should be paused for a second.

An okay-but-not-great solution would be to allow me to specify my sending rate in emails per second in listmonk.

Thank you for maintaining listmonk!

Related issues: #2272 #1332

knadh:
You can throttle and set granular rates in Settings -> Performance.

cmcaine:
Thank you, that's helpful. It's a bit confusing to me that that setting is on the performance tab rather than on the SMTP tab with what look like other connection controls.

I still think it would be nice for one of the other two solutions I mentioned to be included.

Issue #1805, opened 2024-04-02
Sending rate limit


LeoYeAI:
We currently have a Daily Newsletter with one million subscribers. Our goal is to deliver our newsletters simultaneously every morning, ideally within a 5-minute window, assuming SMTP service provider rates allow.

Previously, we have been utilizing Sendy, and despite AWS providing a send rate of 160 emails per second, we struggled to dispatch over 100,000 emails within an hour.

We are considering a transition to Listmonk and would like to know if it can support the delivery of 1-2 million newsletters in the timeframe we require.

Is Listmonk capable of scaling to meet our need for sending a volume of this magnitude? Any implementation advice or insights into the potential performance on a high-capacity workload like ours would be immensely appreciated.

Thank you for your assistance.

knadh:
Hi @LeoYeAI. Our production instance at work has 15+ million subscribers. We use self-hosted instance of Haraka SMTP on EC2. A single instance does about ~50k/minute. More importantly, we don't send marketing e-mails, only a newsletter and rare announcements, hence, the IP reputation with mail providers has been built over many years which permits high throughput sending.

Technically, the rates can be easily horizontally scaled with more Haraka instances. listmonk has no problems scaling to very high throughputs even on small instance, but the bottleneck is going to be:
- The SMTP server(s).
- Mail providers flagging large/fast volumes of e-mails.

LeoYeAI:
Thanks for the real quick response! Knadh.

Issue #1802, opened 2024-03-30
Campaign Finished before all emails Sent


malvibid:
- listmonk: v3.0.0
 - OS: Linux

I need to send a Newsletter to all 1,450 Subscribers in my Newsletter list. I started the campaign, but the campaign finished after sending only 578 / 1,450. Upon investigating I have found there is a 454 Throttling Failure. 

![image](https://github.com/knadh/listmonk/assets/100029192/95b013aa-b364-4d0e-989a-592b9d79cd31)

**Extract of Logs:**
```
2024/03/30 07:05:32start processing campaign (March Newsletter)
2024/03/30 07:05:38error sending message in campaign March Newsletter: subscriber 243: 454 Throttling failure: Maximum sending rate exceeded.
2024/03/30 07:05:38error sending message in campaign March Newsletter: subscriber 268: 454 Throttling failure: Maximum sending rate exceeded.
2024/03/30 07:05:38error sending message in campaign March Newsletter: subscriber 269: 454 Throttling failure: Maximum sending rate exceeded.
2024/03/30 07:05:38error sending message in campaign  March Newsletter: subscriber 252: 454 Throttling failure: Maximum sending rate exceeded.
...
2024/03/30 07:05:55error sending message in campaign March Newsletter: subscriber 1447: 454 Throttling failure: Maximum sending rate exceeded.
2024/03/30 07:05:55campaign (March Newsletter) finished
2024/03/30 07:05:55error sending message 'Finished: March Newsletter': 454 Throttling failure: Maximum sending rate exceeded.

```

My performance settings were the default. 

Concurrency: 10
Message Rate: 10
Batch Size: 10000
Maximum Error Threshold: 1000

I have now adjusted the settings as follows.

![image](https://github.com/knadh/listmonk/assets/100029192/52293c4b-75b5-471d-b6e7-030c2e2b0388)

1.) Would these settings prevent the campaign from Finishing before all the emails are sent? And Pause the campaign and allow me to investigate incase there is an error?

2.) There's ~870 subscribers in this list that didn't receive the email. I want to create a second list to resend the emails to these subscribers only. But how do I retrieve this list? The logs only show me, what I assume is the Subscribers ID. I have extracted these ID's using Excel functions. How can I use these ID's to create a new list - `Newsletter-2`?

MaximilianKohler:
>Would these settings prevent the campaign from Finishing before all the emails are sent?

[...]

PR #2974, opened 2026-03-29
fix: use global rate.Limiter to enforce SMTP send rate across all workers

giform:
## Problem

The current per-worker `message_rate` counter allows the actual SMTP send rate to exceed the configured limit by a factor of `concurrency`. With `concurrency=5` and `message_rate=10`, the effective rate reaches up to 50 msg/sec instead of the expected 10.

The sliding window rate limiter in `pipe.go` only throttles message **enqueueing** into the buffered channel (`campMsgQ`), not the actual SMTP sends. Since the channel buffer is sized `concurrency * message_rate * 2`, workers drain and send messages in bursts that overshoot the configured rate.

This causes SMTP providers with hard rate limits (notably Amazon SES) to return `454 Throttling failure: Maximum sending rate exceeded`, and Listmonk silently skips those subscribers — resulting in campaigns that finish with fewer sends than intended and no way to identify who was missed.

Related issues: #1802, #2273, #645, #119, #1805, #2500

## Fix

Replace the per-worker `numMsg` counter with a single `rate.Limiter` (`golang.org/x/time/rate`) shared across all workers. Each worker calls `rateLimiter.Wait()` right before `messenger.Push()`, ensuring the **actual global send rate** matches the configured limit regardless of concurrency.

The enqueue-side sliding window in `pipe.go` is removed as it is now redundant — rate limiting happens at the point of sending, not enqueueing.

When `sliding_window` is enabled, the rate is derived from `sliding_window_rate / sliding_window_duration`. Otherwise, `message_rate` is used directly as the global per-second limit.

## Changes

- **manager.go**: Add `rate.Limiter` field, initialize from config, replace per-worker `numMsg` with `rateLimiter.Wait()`
- **pipe.go**: Remove sliding window logic from `NextSubscribers()`
- **go.mod**: Add `golang.org/x/time` dependency

## Test results

Tested in production with Amazon SES (14 msg/sec hard limit):

| Setting | Before (bug) | After (fix) |
|---|---|---|
| concurrency=5, rate=12 | 13.9–15.7 msg/sec | **12.1 msg/sec** |
| SES throttling errors | 550 lost in a 3726-subscriber campaign | **0** |
| Campaign completion | 85% delivered | **100%** |

## Breaking changes

None. Existing config values (`message_rate`, `sliding_window_rate`) are respected — the only change is that they now work as users expect (global limit, not per-worker).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,9 +2,19 @@ frontend/node_modules/
 frontend/.cache/
 frontend/yarn.lock
 frontend/build/
+frontend/public/static/email-builder/
+frontend/dist/
+frontend/email-builder/dist/
+email-builder/node_modules/
+email-builder/.cache/
+email-builder/yarn.lock
+email-builder/dist/
+static/public/static/altcha.umd.js
 .vscode/
 
 config.toml
+docker-compose.override.yml
 node_modules
 listmonk
 dist/*
+uploads/
--- a/go.mod
+++ b/go.mod
@@ -68,7 +68,7 @@ require (
 	golang.org/x/net v0.23.0 // indirect
 	golang.org/x/sys v0.18.0 // indirect
 	golang.org/x/text v0.16.0 // indirect
-	golang.org/x/time v0.5.0 // indirect
+	golang.org/x/time v0.5.0
 	google.golang.org/appengine v1.6.8 // indirect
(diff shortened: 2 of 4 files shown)


knadh:
Hi @giform. That's numerous breaking issues. Please reopen once you've been able to thoroughly verify at your end.

Issue #2576, opened 2025-07-15
Outlook design issues


jiimmyy1:
Hi,

I'm using the visual editor and I'm having the typical width issue with the text and images.
I have seen lots of opened issues but I don't find a solution when using visual editor.
Can somebody help?

Thanks

MaximilianKohler:
Not enough details. Post screenshots.

jiimmyy1:
<img width="1875" height="762" alt="Image" src="https://github.com/user-attachments/assets/fefc2a1f-fcea-4b3d-9f98-eec65805e4f3" />

jiimmyy1:
Please any solution with this. I think its very important to solve this but I dont see any guides with visual editor.

knadh:
I'm not familiar with Outlook or its rendering quirks, unfortunately. Hoping that someone who's familiar with Outlook will share insights here.

andi98:
Hi, we had the same issue: you have to ensure that your images have max. 600px width. Then it looks fine (for us...)

jiimmyy1:
I will give a try. What about height? Max 600px? Auto? Thanks

candidexmedia:
@jiimmyy1 That's very odd... I tested it on my own Listmonk instance and the Visual editor, with an image that was 1080p wide, and didn't see any issues with my Outlook Android app.

I even tested with the image set to auto width, then 5000px width.

Could you let us know which Outlook app you are using and the version?

Also, if you're comfortable doing so: could you also share a json export of your message for folks to try?

jiimmyy1:
Hi, the issues comes when using Outlook on PC. Not sure if also in macOS.

Waiting for @andi98 answer to my comment for test. Thanks!

MrJoky:
MS Outlook requires special VML markup to render certain elements correctly, such as images and buttons.
In addition, Outlook relies on fixed widths for elements; otherwise, the layout may be displayed incorrectly.
From what I have seen, the buttons already include part of this setup, but the VML portion is still missing. The images, however, appear to be set up incorrectly altogether.
I am not sure whether I am allowed to post two links here that show how images and buttons can be implemented safely for Outlook.

candidexmedia:
@MrJoky You can share links on GitHub, yes.

You can also format code using the following symbol: ` ``` `

MrJoky:
```
<td background="https://i.imgur.com/YJOX1PC.png" bgcolor="#7bceeb" width="600px" height="200px" valign="top">
[...]

Issue #1231, opened 2023-03-06
Outlook issues


yourics:
I'm having issues with the formatting in MS Outlook.
For some reason a lot of things are shifted, full width or the e-mail will display like on a phone, while being on a desktop PC.
Everywhere my e-mail displays correctly, except for MS Outlook..

Could somebody PLEASE help me? I have no idea where to start solving this problem.
I've already edited my template so it only contains {{ template "content" . }} as the templating is done in MJML. With the default template it was even worse, so that's why I removed everything except for this one line (of course).
Also is there a way to bypass the dark mode? Some of the colors are displayed incorrectly as you can see.

I've included my .mjml file as well as the generated HTML file by MJML: 
[Nieuwsbrief Maart 2023.zip](https://github.com/knadh/listmonk/files/10897495/Nieuwsbrief.Maart.2023.zip)

Here are some examples of the correct formatting on outlook web and the wrong formatting on the MS Outlook application.
Correct: https://gyazo.com/9fc848284e5ec57d3bcffbf88d3ac893

Wrong: https://gyazo.com/b2787e9dcbc4072940b803108b035629

yourics:
Anyone has any idea how to fix this? I have a newsletter that I need to send ASAP. Any help or reply would be appreciated

candidexmedia:
Outlook is notorious for messing up layouts, and the experience will vary between Outlook desktop, Outlook online, Outlook on certain OSs, etc.

Short-term solution: Your best bet would be to include a "View in browser" link so that Outlook users can view the correct version. Do you know for a fact that a significant portion of your recipients use Outlook?

Long-term: you may need to connect with the MJML community to figure out what the issue is. A quick Google search of "MJML Outlook" shows that others are having issues with getting the framework to work with Outlook:

![image](https://user-images.githubusercontent.com/55474996/225984221-5ce3fa6e-2a88-4d90-be3d-9e2d07ab11ed.png)

You can also review the compatibility of the MJML components you used with major email clients: https://mjml.io/faq#email-clients

PR #2978, opened 2026-03-30
 Outlook-safe rendering for visual email builder output

MrJoky:
This PR improves Outlook compatibility for emails created with the visual email builder.

  It adds Outlook-oriented post-processing to the visual builder HTML output so that buttons, images, text blocks, and
  spacers render more reliably in Outlook. It also fixes format switching in the campaign editor so that switching
  between `visual` and `raw/html` does not discard the visual builder source for unchanged content.
  
  References: #2576 
  
  Not sure if this issue addresses the same problem #1231


knadh:
Thanks @MrJoky. This is a big PR, which is going to be very difficult for me to test and verify. Would you be able to share screenshots+HTML of e-mails post-processed via this rendering in Outlook and non-Outlook clients?

MrJoky:
@knadh [Different clients](https://testi.at/proj/692p04zo0la4h5e6u9o).

MrJoky:
@knadh how far are you with the review?

MrJoky:
> ### ssues Found
> 
> **Critical (P0/P1)**
> 
>     * **[P1] Fix malformed  tag injection when no  exists** (`frontend/email-builder/src/utils.tsx:13-18`)
>       
>       * `injectHeadContents()` replaces the opening `<html ...>` tag with `<html$1><head>...` but misses the closing `>` after the `<html...` portion, producing invalid markup like `<html<head>...` (or `<html lang="en"<head>...`). This path is taken whenever the renderer output has an `<html>` but no `<head>` (which is the common case in this builder), so `renderHtmlWithMeta(document, {rootBlockId: 'root'})` can emit broken HTML and downstream clients/fallback paths will persist or display malformed content.
> 
> 
> ### Summary
> 
> Total issues: 1 critical, 0 important, 0 minor.
> ### Overall Verdict
> 
> **Status**: Patch has blocking issues
> 
> **Explanation**: The new head-injection helper can generate malformed HTML (`<html<head>`) when the input lacks a `<head>` tag, which is likely for the visual builder output when `outlook` is not enabled. This can break rendered/saved email HTML in production paths that call `renderHtmlWithMeta` without the `outlook` option.

False positive: current implementation already emits `<html...><head>...`, not `<html<head>`. Verified against `frontend/email-builder/src/utils.tsx` and with a local runtime check using `<html lang="en"><body>...</body></html>` input.

MrJoky:
@knadh all issues from hodor are solved, except last one which is not correct.

MrJoky:
https://testi.at/proj/bjkrf423cz67fdldio Current cross-platform test

knadh:
Thanks @MrJoky.

I do find it wrong to pollute all markup for all clients just for outlook compatibility (which likely is a tiny % of recipients) :)

I've pushed a commit that makes it optional by adding a UI toggle to the global pane in the editor. Makes sense?

image

MrJoky:
> Thanks @MrJoky.
> 
> I do find it wrong to pollute all markup for all clients just for outlook compatibility (which likely is a tiny % of recipients) :)
> 
> I've pushed a commit that makes it optional by adding a UI toggle to the global pane in the editor. Makes sense?
> image

AH! That's an awesome idea! I like it.

PR #2996, opened 2026-04-10
Fixs #2987 - Prevent confirmed subscriber becoming unconfirmed when re-subscribing.

### LINKED ISSUE #2987: Subscribing to double opt-in list after confirming subscription changes status to unconfirmed
**Version:**
 - listmonk: nightly-2026-04-01
 - OS: Docker container

1. Create a public list with double opt-in.
2. Subscribe a user to the list.
3. Confirm subscription.
4. Re enter email to subscribe to the list using form (even though already subscribed and confirmed).
5. The user is greeted with a "An e-mail has been sent to you to confirm your subscription(s)." message. The subscription status is now changed to "unconfirmed" and new confirmation email is sent.

Expected result: User should still be greeted with "An e-mail has been sent to you to confirm your subscription(s)." to avoid leaking information about subscription status, but the email shouldn't be marked as "unconfirmed" and no new confirmation email should be sent.

**Discussion**
This is a minor bug. As it stands this doesn't leak information about subscription status. However it does make it possible for unauthorised users to effectively unsubscribe an email from a double opt-in list with just the target email address.

ref: 
- #2976
- b0da6bd

blu3id:
This closes #2987 or at least fixes the fix of the regression identified in #2976

A confirmed double opt-in subscriber is no-longer unsubscribed if they re-subscribe to the list (while still subscribed and confirmed).

This doesn't address the discussion point of not leaking information about subscriber status as a confirmed subscriber on a double opt-in list will now get the message "Subscribed successfully." rather than "An e-mail has been sent to you to confirm your subscription(s)." to avoid leaking information about subscription status". 

However this was how things have always been and would be a larger piece of work/ restructuring. A work around is adjusting the messages in the i18n language file to remove the difference between the messages.

### DIFF
--- a/queries/subscribers.sql
+++ b/queries/subscribers.sql
@@ -189,11 +189,12 @@ INSERT INTO subscriber_lists (subscriber_id, list_id, status)
     SET status = (
         CASE
             WHEN $4='blocklisted' THEN 'unsubscribed'::subscription_status
-            -- When $11 (allow resubscribe) is true, override existing statuses (used by public subscription form).
+            -- When $11 (allow resubscribe) is true, override existing statuses except confirmed (used by
+            -- public subscription form).
+            WHEN subscriber_lists.status = 'confirmed' THEN 'confirmed'
             WHEN $11 = TRUE THEN $8::subscription_status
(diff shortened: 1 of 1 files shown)


blu3id:
Comment adjusted per review.

Issue #2580, opened 2025-07-17
Azure smtp webhook support for bounce


RogerioTeixeira:
**Is your feature request related to a problem? Please describe.**

I'd like to implement support for Azure Communication Services email webhook events.

**Describe the solution you'd like**
Azure supports posting events such as Delivered, Bounced, Opened, etc., as JSON via webhooks.
My plan would be to:

Add a new webhook handler under /api/webhooks/azure

Store delivery or bounce info in the DB

PR #3001, opened 2026-04-12
Add Azure ACS bounce webhooks and settings support

oskari:
Azure ACS is not the most sophisticated platform, but I managed to make it work. Azure documentation around this is a bit lacking (especially Message-ID -> internetMessageId part) but at least on my tests this is working. 

Anything I should do differently here or anything that I should clarify?


oskari:
Just to document, this solves #2580

internal/messenger/email/email.go
```diff
@@ -156,6 +158,18 @@ func (e *Emailer) Push(m models.Message) error {
 		em.Headers.Set(k, v[0])
 	}
 
+	// Azure Event Grid preserves Message-ID. Generate one only for Azure SMTP
+	// when upstream headers have not set one explicitly.
+	if isAzureSMTPHost(srv.Host) && em.Headers.Get(models.EmailHeaderMessageId) == "" {
```
knadh: This is not clean, unfortunately, hardcoding Azure specific logic dependent on its domain names in the generic e-mail flow.

oskari: It's not ideal yeah. Now that 78f6af0 is implemented, maybe I could extend it to include campaign and subscriber id encoded in it?

knadh: hmm, encoding parseable semantic values, that too IDs, in the `Message-Id` value seems problematic to be honest. It's strange that Azure strips all headers. I'm unable to think of a solution right now, but ascribing semantic value to `Message-Id` globally just to support Azure doesn't feel right.

oskari: I understand the feeling. 

As long as the Message-Id is RFC compliant and we use the local part (Which this doesn't, but that's a quick fix), I don't see it being too bad. It could easily be campaign-id+subscription-id+timestamp@domain and be RFC compliant and unique. The data is transmitted in the headers anyway, so it's there. I'd rather keep it azure-specific, but understand the will to keep the email-flow generic.

Azure is what it is. I've explored all the options but the event grid schema seems to be what it is and the only "answer" in general seems to be: Use Message-Id. 

Two options come to my mind, if we don't want to go this route:
1. Keep random message-id, don't correlate the events to campaigns. We can still match emails from the event to subscribers. Just not to campaigns.
2. Save the random message-id to db, lookup during webhook event, truncate data after x days. Might be problematic on high volume

Unfortunately even using the ACS api to send mails won't help here, as the problem is event grid side.

Let me know how you feel about these options

knadh: I'd prefer going with 1) and not making message ID semantic for now (even if not being able to detect campaigns is not ideal). Storing message IDs in the DB for lookups just to support Azure is a bad idea.

i18n/en.json
```diff
@@ -411,6 +412,10 @@
     "settings.bounces.scanInterval": "Scan interval",
     "settings.bounces.scanIntervalHelp": "Interval at which the bounce mailbox should be scanned for bounces (s for second, m for minute).",
     "settings.bounces.sendgridKey": "SendGrid Key",
+    "settings.bounces.azureSharedSecret": "Azure Event Grid Shared Secret",
+    "settings.bounces.azureSharedSecretHelp": "Provide the shared secret configured for your Azure Event Grid webhook endpoint. listmonk accepts the secret via query parameter ?code=... or via the configured header name.",
```
knadh: `listmonk accepts the secret via query parameter ?code=... or via the configured header name.`

I think this is unnecessary technical detail in the UI.

PR #3015, opened 2026-04-22
fix: allow creating campaign drafts without lists assigned

### LINKED ISSUE #3014: Campaign content editor disabled when no list is assigned on new campaign (non-admin user)
**Version:**
 - listmonk: [v6.1.0]
 - OS: [Docker]

When creating a new campaign as a non-admin user with `campaigns:manage` permission, the content editor is disabled/non-functional until a list is selected in the campaign settings.

Steps to reproduce:
1. Create a user role with `campaigns:manage` (without `lists:get_all`) and assign a list role with at least one list.
2. Log in as that user and create a new campaign.
3. Without selecting a list, navigate to the content editor tab.
4. The editor is disabled and content cannot be saved.

**Expected behavior:**
The content editor should be usable regardless of whether a list has been assigned yet. Selecting a list and writing campaign content are independent steps in the campaign creation flow.

**Actual behavior:**
The content editor is blocked until a list is assigned to the campaign. This forces an unnatural workflow and is confusing for non-admin users.

**Screenshots:**

Image

Danimarqz:
Lists are now only required when starting or scheduling a campaign, not when saving a draft. This unblocks the content editor tab for non-admin users who previously could not save a campaign (and thus access the editor) without first selecting a list.

Fixes #3014

### DIFF
--- a/cmd/campaigns.go
+++ b/cmd/campaigns.go
@@ -365,6 +365,17 @@ func (a *App) UpdateCampaignStatus(c echo.Context) error {
 		return err
 	}
 
+	// Lists are only required when starting or scheduling a campaign, not for drafts.
+	if req.Status == models.CampaignStatusRunning || req.Status == models.CampaignStatusScheduled {
+			camp, err := a.core.GetCampaign(id, "", "")
+			if err != nil {
+					return err
+			}
+			if len(camp.Lists) == 0 {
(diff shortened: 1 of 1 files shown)


filipesmedeiros:
My opiníon regarding the AI's review is that no-list campaigns can be accessible to anyone (who can manage all lists). For people that have only list-based permissions, I think they should not be able to manage these lists.

Also, if a campaign is scheduled, it should not be possible to remove all lists from it. First you should have to change it back to draft.

PR #3039, opened 2026-05-12
fix(manager): pipe cleanup race + interruptible window sleep + list-view click dedup

solomon-the-genius:
## Summary

Two recurring root-cause bugs in the Solomon fork, plus updated operator help text on the rate-limit setting. Replaces band-aid fixes with the underlying mechanics that should keep campaigns from getting stuck or showing inconsistent numbers.

## 1. Pipe cleanup race (Friday May 8 -> Monday May 11 stall)

A pipe hits the sliding-window cap, sleeps for up to `SlidingWindowDuration`, gets paused mid-sleep, and on resume `scanCampaigns` installs a NEW pipe at `m.pipes[id]`. The old goroutine eventually wakes, runs deferred cleanup, and unconditionally `delete(m.pipes, id)` — which removes the **successor** pipe. Campaign goes silent until the container restarts.

Three changes in `internal/manager/pipe.go`:
- `pipe.cleanup` defer now does a pointer-identity check (`m.pipes[id] == p`) before deleting.
- `Stop()` closes a per-pipe `done` channel (sync.Once), so a goroutine parked in the window-cap sleep wakes immediately on pause/cancel.
- `newPipe` stops any pre-existing pipe at `m.pipes[id]` before installing itself, belt-and-suspenders for evergreen re-queue churn.

## 2. Click count divergence (22 list view vs 8 analytics)

`get-campaign-stats` hardcoded `COUNT(campaign_id)` on `link_clicks` / `campaign_views`, while `get-campaign-analytics-unique-counts` dedupes per subscriber when `privacy.individual_tracking` is on. List view always showed raw event counts.

- `queries/campaigns.sql`: views/clicks CTEs now take a `%s` count expression.
- `cmd/init.go`: `prepareQueries` swaps in `COUNT(campaign_id)` or `COUNT(DISTINCT subscriber_id)` at boot, mirroring the analytics path.

## 3. Operator help text on the rate-limit setting

`i18n/en.json`: rewrote `slidingWindowHelp` / `slidingWindowRateHelp` so the operator can change the cap themselves without asking. Makes clear it's per-campaign, what happens at the cap (sleep until window expires), and that 50 means 50/hr per campaign with the default 1h window.

## Test plan

- [ ] Build green (Docker make build)
- [ ] Pause a campaign mid-sleep (after 50-msg burst), wait 30 minutes, resume; sends restart within ~1 minute and pipe map is clean
- [ ] List view click count == analytics click count for any campaign with multi-click subscribers
- [ ] Settings -> Performance shows new help text under the sliding-window field

🤖 Generated with [Claude Code](https://claude.com/claude-code)


solomon-the-genius:
Wrong target — meant for our internal fork repo, not upstream. Closing.

PR #3056, opened 2026-05-19
Allow Go templates in campaign From header

nemunaire:
Some concurrent softwares (like ActiveCampaign) allows template in the From header.

This pull request adds support for this!


knadh:
Hi @nemunaire. I refactored and simplified the logic and added it to the central template compilation, rather than it being a separate `if` condition in the messenger. This is cleaner and consistent with all other template compilation logic in general (altbody, headers etc.). Please take a look (I've tested it, but please do at your end as well) and let me know.

nemunaire:
Hi @knadh. Tested on top of v0.6.2. It also works for me as expected. :rocket:

PR #3057, opened 2026-05-19
New bounces source: Oracle Cloud Infrastructure suppression-list

nemunaire:
This pull request adds support for the Oracle Cloud Infrastructure suppression list. OCI doesn't provide any kind of webhook, so we need to poll the account regularly instead.

image

This follows the same pattern as the mailbox poller.

If Listmonk is the only software sending email through OCI, it's safe to delete items from the suppression list. Otherwise, the poller handles deduplication to avoid reporting the same bounce multiple times. A boolean can be set by the user to choose the wanted behavior.

The OCI request-signing logic has been reimplemented here to avoid adding a dependency on the huge OCI SDK.


knadh:
Thanks for the PR @nemunaire, but I don't think it makes sense to merge this into the core. It's better to run this as a side-car service that polls Oracle cloud, then communicates with listmonk via APIs.

Issue #3076, opened 2026-06-05
`@TrackLink` Corrupts Tracked Links on Complex URLs


gregouzdev:
**Version:**
 - listmonk: v6
 - OS: Docker

## Affected scope

- Module: email campaigns
- Feature: link tracking
- Affected syntax: `@TrackLink`
- Affected link types:
  - Platform URLs
  - Long URLs
  - URLs containing complex query parameters
  - URLs containing special characters
  - URLs containing Unicode characters
  - URLs containing emojis
- Observed impact:
  - Sent links become invalid;
  - Client links are no longer clickable properly;
  - The `@TrackLink` suffix may be retained or misinterpreted in the final URL;
  - Sent campaigns may contain corrupted links.

## Root cause

The issue is located in `./models/common.go`, in the regex pattern used to detect `@TrackLink` occurrences.

The current pattern strictly follows **RFC 3986** URL syntax, which excludes characters such as Unicode, emojis, and unencoded special characters. However, real-world client URLs — particularly those generated by platforms like SharePoint — often follow **IRI (Internationalized Resource Identifier)** conventions, which allow a broader character set.

As a result, the regex fails to match these URLs and leaves the `@TrackLink` suffix unprocessed and appended to the raw URL.

### Suggested fix

Replacing the current pattern with a more permissive one that accommodates IRI-style URLs:

```go
(https?://[^\s<>"{}]*)@TrackLink\b
```

This pattern captures everything up to the first whitespace or structural HTML character, making it robust against URLs containing Unicode characters, emojis, encoded sequences, and long query strings.

### Basic example

```html
<a href="https://client.platform.com/sites/Project/Documents%20shared/Client%20form%20✅.aspx?web=1&param=value@TrackLink">
  Open client document
</a>
```

### Long URL example

```html
<a href="https://client.platform.com/:x:/r/sites/Management/Shared%20Documents/Reports/2026/Marketing%20campaign%20report%20-%20final%20version%20📊.xlsx?d=w123456789abcdef&csf=1&web=1&e=AbCdEf@TrackLink">
  View report
</a>
```

### Special characters example

```html
<a href="https://client.example.com/document?id=12345&name=client-presentation&status=validated&emoji=✅@TrackLink">
  View presentation
</a>
```

### Partially encoded URL example

```html
<a href="https://client.platform.com/sites/Project/Shared%20Documents/Team%20meeting%20notes%20🚀.docx?web=1&download=1@TrackLink">
  Download meeting notes
[...]

PR #3077, opened 2026-06-05
feat: add switch RFC3986 / IRI-Style for @TrackLink

gregouzdev:
Response for Issue #3076 

## Summary

The root cause was identified in `models/common.go` (line 55): the existing regex pattern only accepted a strict subset of RFC 3986 characters (ASCII letters, digits, and a limited set of symbols). As a result, URLs containing emojis or non-ASCII Unicode characters caused the pattern to break before reaching `@TrackLink`, preventing the shortcut from being converted into the expected `{{ TrackLink ... }}` template tag — leaving the raw, corrupted URL in the final email output.

### `models/common.go` (line 45) — configurable regex mode via `LISTMONK_TRACKLINK_REGEX_MODE`

The fix introduces an environment variable to control the URL matching strategy, allowing operators to choose between improved compatibility and strict legacy behaviour:

| Mode | Behaviour |
|------|-----------|
| `iri` *(default)* | Captures URLs up to common HTML/template delimiters (`"`, `'`, whitespace, `<`, `>`, `{`, `}`) — handles emojis, Unicode, encoded sequences, and long query strings |
| `unicode` | Restores the previous strict regex for environments that require it |

### Regex comparison

**Before (strict RFC 3986):**
```go
// Only matched ASCII letters, digits, and a limited set of symbols
// — broke on emojis and non-ASCII Unicode characters
```

**After (IRI-style, default):**
```go
(https?://[^\s<>"'{}]*)@TrackLink\b
```

This pattern captures everything up to the first structural HTML or template delimiter, making it robust against the full range of real-world client URLs.

## Backward compatibility

This change is **non-breaking**. The new `iri` mode is the default, but the previous behaviour can be fully restored by setting:

```env
LISTMONK_TRACKLINK_REGEX_MODE=unicode
```

## Testing

- [x] URLs with emojis (e.g. `✅`, `📊`, `🚀`)
- [x] URLs with non-ASCII Unicode characters (e.g. accented characters, CJK)
- [x] URLs with partial encoding (`%20`, `%2F`, etc.)
- [x] Long URLs with complex query strings
- [x] Standard ASCII URLs — no regression
- [x] Legacy `unicode` mode — behaviour identical to previous implementation

### DIFF
--- a/models/common.go
+++ b/models/common.go
@@ -4,7 +4,9 @@ import (
 	"database/sql/driver"
 	"encoding/json"
 	"fmt"
+	"os"
 	"regexp"
+	"strings"
 
 	"github.com/yuin/goldmark"
 	"github.com/yuin/goldmark/extension"
@@ -40,6 +42,16 @@ type regTplFunc struct {
 	replace string
 }
 
+const trackLinkRegexModeEnv = "LISTMONK_TRACKLINK_REGEX_MODE"
+
+func trackLinkShorthandRegExp() *regexp.Regexp {
+	if strings.EqualFold(os.Getenv(trackLinkRegexModeEnv), "unicode") {
+		return regexp.MustCompile(`(https?://[\p{L}\p{N}_\-\.~!#$&'()*+,/:;=?@\[\]%]*)@TrackLink`)
+	}
+
+	return regexp.MustCompile(`(https?://[^"'\s<>{}]+)@TrackLink`)
+}
+
 var regTplFuncs = []regTplFunc{
 	// Regular expression for matching {{ TrackLink "http://link.com" }} in the template
 	// and substituting it with {{ TrackLink "http://link.com" . }} (the dot context)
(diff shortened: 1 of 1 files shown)


knadh:
There should be no `os.Getenv()`-level switch. This should be handled solely in the regexp.

gregouzdev:
Good point regarding the environment switch. I agree this can be handled directly in the regexp and doesn't need an additional configuration option.

The underlying issue is that `@TrackLink` is not really validating URLs, but extracting the URL portion before transforming it into `{{ TrackLink ... }}`.

The current character-class approach requires continuously enumerating allowed URL characters. Using a delimiter-based approach is simpler because `@TrackLink` only needs to identify where the URL ends, not validate which characters are allowed inside it.

Furthermore, in real-world conditions, some platforms like SharePoint or similar systems generate very long and complex URLs (parameters, nested paths, tokens, etc.), making strict or overly restrictive approaches even less suitable.

Rather than managing multiple variants, it's possible to simplify by using a single, more permissive regular expression that works correctly: `(https?://[^"'\s<>{}]+)@TrackLink`

PR #3080, opened 2026-06-08
fix: TrackLink Unicode/emoji URLs, bounce multipart infinite loop, opt-in List-Unsubscribe

### LINKED ISSUE #3063: `List-Unsubscribe` header broken on double opt-in confirmation emails
**Version:**
 - listmonk: v6.1.0
 - OS: Docker

1. Ensure ``Include `List-Unsubscribe` header`` setting is enabled.
2. Create a public double opt-in list.
3. Subscribe a user to the list.
4. Issue a `POST` request to the `List-Unsubscribe` header in the email sent to the user (What the `List-Unsubscribe-Post: List-Unsubscribe=One-Click` tells mail clients to do).

_Expected result:_ User is marked as unsubscribed for the unconfirmed list(s) or blocklisted?
_What actually happens:_ Nothing - if you examine the response to the `POST` you see a "Unsubscribed successfully" message.

**Discussion:**

I came across this accidentally exploring the code and looking closely at headers etc. to investigate #3037. The issue is the unsubscribe URL is populated as follows using the "dummy" UUID: `/subscription/00000000-0000-0000-0000-000000000000/{subscriber-uuid}` this means that the POST handler is unable to identify the list to unsubscribe the user from.

While this "bug" doesn't pose a big issue it technically is abusing the `List-Unsubscribe` header standard as the "one click" doesn't do anything. Though the argument can be made that the user hasn't confirmed their subscription so there is nothing to unsubscribe from.

I don't know what the best approach to fixing this would be options I can think of:
1. Blocklist the user (though could have unintended behaviour as would unsubscribe from other lists) 
2. Mark the user as unsubscribed for all unconfirmed lists in the opt-in email
3. Don't include the header in opt-in confirmation emails
4. Leave as is

### LINKED ISSUE #3071: error reading multipart bounce message: multipart: NextPart: EOF
**Version:**
 - listmonk:  v6.1.0
 - OS: linux (ubuntu)

My listmonk log file is full of messages like this:
`2026/06/02 21:18:43.435411 pop.go:133: error reading multipart bounce message 2: multipart: NextPart: EOF`

This looks like a duplicate of issue #1848, but it seems to be closed as won't fix and is not receiving any attention.

I'd love to provide an example bounce email that causes this error, but I can't find the offending one among the hundreds of bounces I have. 
They all look like they have a body, so I'm not sure what the issue is. If the log listed some message ID I could use for searching, that would help a lot.

### LINKED ISSUE #3076: `@TrackLink` Corrupts Tracked Links on Complex URLs

buddywhitman:
Fixes three reported bugs with no existing open PRs.

  ## Changes
  
  ### Fix #3076 — TrackLink regex drops URLs containing emoji or non-ASCII Unicode
  
  The original regex `[\p{L}\p{N}_\-\.~!#$&'()*+,/:;=?@\[\]%]*` covers Unicode
  letters/numbers but not emoji (Unicode category `So`), so campaign links with
  emoji in query params or fragments were silently truncated.

  **Fix:** replace with `\S+?` (lazy non-whitespace). The `@TrackLink` suffix anchors
  the match, so the lazy quantifier stops at the right boundary and handles any
  non-whitespace URL character including emoji.
  
  ### Fix #3071 — bounce POP processor: infinite loop on broken multipart message
  
  In `internal/bounce/mailbox/pop.go`, the multipart-parse loop used `continue`
  on non-EOF errors from `mr.NextPart()`. A broken reader returns the same error
  forever, spinning the loop indefinitely.

  **Fix:** `break` instead of `continue` on non-EOF errors.
  
  ### Fix #3063 — List-Unsubscribe header non-functional on opt-in (non-campaign) emails
  
  Opt-in emails use `dummyUUID` as the campaign UUID in the unsubscribe URL, which
  the route rejects — so clicking the header link returned a 404/error. The code
  also set `List-Unsubscribe-Post` (RFC 8058 one-click), which requires the URL to
  actually work.

  **Fix:** Remove `List-Unsubscribe-Post` (one-click requires a working endpoint).
  Keep the informational `List-Unsubscribe` header using the already-resolved
  `out.UnsubURL` rather than re-constructing it with `dummyUUID`.

  ## Testing
  - `go build ./...` passes
  - Manually traced each code path against the issue reproduction steps

### DIFF
--- a/cmd/subscribers.go
+++ b/cmd/subscribers.go
@@ -871,11 +871,14 @@ func makeOptinNotifyHook(unsubHeader bool, u *UrlConfig, q *models.Queries, i *i
 		hdr := textproto.MIMEHeader{}
 		hdr.Set(models.EmailHeaderSubscriberUUID, sub.UUID)
 
-		// Attach List-Unsubscribe headers?
+		// Attach List-Unsubscribe header for opt-in emails.
+		// Note: List-Unsubscribe-Post (RFC 8058 one-click) is intentionally
+		// omitted here because opt-in emails use a dummy campaign UUID in the
+		// unsub URL, which the unsubscribe route rejects. The informational
+		// List-Unsubscribe header is still set so that email clients display a
+		// visible unsubscribe link that the subscriber can follow manually.
 		if unsubHeader {
-			unsubURL := fmt.Sprintf(u.UnsubURL, dummyUUID, sub.UUID)
-			hdr.Set("List-Unsubscribe-Post", "List-Unsubscribe=One-Click")
-			hdr.Set("List-Unsubscribe", `<`+unsubURL+`>`)
+			hdr.Set("List-Unsubscribe", "<"+out.UnsubURL+">")
 		}
 
 		// Send the e-mail.
--- a/internal/bounce/mailbox/pop.go
+++ b/internal/bounce/mailbox/pop.go
@@ -130,8 +130,12 @@ func (p *POP) Scan(limit int, ch chan models.Bounce) error {
 				if err == io.EOF {
 					break
 				} else if err != nil {
+					// A non-EOF error means the multipart reader is in a broken
+					// state (e.g. malformed boundary). Continuing the loop would
+					// call NextPart() on the same broken reader forever. Break
+					// and process whatever we have so far.
 					p.lo.Printf("error reading multipart bounce message %d: %v", id, err)
-					continue
+					break
 				}
 				h = part
 			}
--- a/models/common.go
+++ b/models/common.go
@@ -52,10 +52,11 @@ var regTplFuncs = []regTplFunc{
 	// Convert the shorthand https://google.com@TrackLink to {{ TrackLink ... }}.
 	// This is for WYSIWYG editors that encode and break quotes {{ "" }} when inserted
 	// inside <a href="{{ TrackLink "https://these-quotes-break" }}>.
-	// The regex matches all characters that may occur in an URL
-	// (see "2. Characters" in RFC3986: https://www.ietf.org/rfc/rfc3986.txt)
+	// Uses \S+? (lazy non-whitespace) to capture any URL character including Unicode,
+	// emoji (e.g. example.com/path/👍), and IRI characters beyond RFC3986 ASCII.
+	// The lazy quantifier stops at the first @TrackLink suffix, not earlier.
 	{
-		regExp:  regexp.MustCompile(`(https?://[\p{L}\p{N}_\-\.~!#$&'()*+,/:;=?@\[\]%]*)@TrackLink`),
+		regExp:  regexp.MustCompile(`(https?://\S+?)@TrackLink`),
 		replace: `{{ TrackLink "$1" . }}`,
 	},
 


blu3id:
Hi @buddywhitman,

Thanks for the contributions. Regarding #3063 (the issue I created) I don't think this is a helpful fix. While technically correct removing the `List-Unsubscribe-Post` header means the `List-Unsubscribe` is no-longer "broken" per RFC8058 all the major MSPs (Gmail, Microsoft, etc.) expect both a `List-Unsubscribe` and `List-Unsubscribe-Post` header to meet their bulk mail requirements for easy unsubscribing.[^1] This is particularly important given the [discussion](https://github.com/knadh/listmonk/issues/3063#issuecomment-4545725534) in the issue regarding the reason for why this header was added to the opt-in emails in the first place. It would have been good to discuss this in the issue (which I thought made it clear that the technical solutions was still up for discussion) before opening a PR.

[^1]: https://support.google.com/a/answer/81126?hl=en

blu3id:
On a broader but linked note to the above it isn't great practice to lump multiple issues and their fixes into a single PR as it makes discussion and eventual merging harder for maintainers (FYI which I am not for this project)

> Fixes three reported bugs with no existing open PRs.

#3076 has a PR that is already open - #3077 which was opened ~3 days ago
#3071 has a PR that is already open - #3075 which was opened ~3 days ago

buddywhitman:
Hi @blu3id 
Thank you for your feedback! My bad for not discussing the implementation before opening a PR, and thank you for pointing out the newly opened PRs for the other issues. I implemented these fixes a couple of days back, but only created a PR yesterday. Well, regardless, the superseding fixes have superior implementation.

Have some ideas for implementing a fix for your flagged issue, will discuss further in the respective issue thread. Closing this PR for now.

PR #3091, opened 2026-06-16
Show the "Links" chart heading on the campaign analytics page

### LINKED ISSUE #3089: Campaign analytics: the "Links" chart has no heading
On the campaign analytics page every chart has an `<h4>` heading (Views, Clicks, Bounces) except the **Links** bar chart, which renders with no title.

It's a leftover from the c3.js→chart.js migration: the `links` chart config still carries `chart: null`, which the heading's `v-if="v.chart !== null"` keys off. In the c3.js code that property held the chart instance and the gate meant "show the heading once rendered"; after the refactor it became vestigial and was removed from the other charts but kept on `links`, leaving it the only chart without a heading.

Expected: the Links chart shows its heading like the others.

sysrow:
Fixes #3089.

The Links bar chart was the only analytics chart without a heading because its config still carried `chart: null` purely to gate the `<h4>` — a leftover from the c3.js→chart.js migration. This drops the dead property and renders the heading like the other charts.

The count parenthetical is omitted for the Links (bar) chart: its only available total is the sum of per-link clicks, which duplicates the Clicks chart's total, and the per-link counts are already shown on the bars.

### DIFF
--- a/frontend/src/views/CampaignAnalytics.vue
+++ b/frontend/src/views/CampaignAnalytics.vue
@@ -54,9 +54,9 @@
         <div class="columns">
           <div class="column is-9">
             <b-loading v-if="v.loading" :active="v.loading" :is-full-page="false" />
-            <h4 v-if="v.chart !== null">
+            <h4>
               {{ v.name }}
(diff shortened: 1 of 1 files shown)


sysrow:
firefox_3nzgYCSnoR

Issue #2145, opened 2024-11-09
Support Redirect After Form Submission 


hay-kot:
**Is your feature request related to a problem? Please describe.**

Not really a problem as much as a limitation of the current behavior. 

**Describe the solution you'd like**

When creating a form for listmonk I would like the ability to specify a redirect URL to send the user to instead of showing them the listmonk page. For example, Convert Kit allows a user to create a redirect link, or show a success message in their form builder. 

![CleanShot 2024-11-09 at 12 12 40@2x](https://github.com/user-attachments/assets/ff7b7c85-c15a-439a-a7e0-bf2413a6509e)

I would love it if I could add a hidden input to manage a redirect to a custom page after submission. 

```html
<form method="post" action="..." class="listmonk-form">
          <div>
          <h3>Subscribe</h3>
          <input type="hidden" name="nonce" />
          
          <p><input type="email" name="email" required placeholder="E-mail" /></p>
          <p><input type="text" name="name" placeholder="Name (optional)" /></p>
      
        <p>
          <input id="60e8f" type="checkbox" name="l" checked value="60e8f519-972b-43b9-a92f-0ee72e32ff89" />
          <label for="60e8f">Launch List</label><br /><span>Launch List</span>
        </p>

        <!-- Redirect setting/definition --!>
        <input name="redirect_url" value="https://example.com/" type="hidden"/>
        
        <p><input type="submit" value="Subscribe" /></p>
    </div>
</form>
```

Then if `redirect_url` is found in the form fields, you'd be redirected to whatever value that is.

candidexmedia:
Is this for a standalong form page, or form embed on a website?

hay-kot:
My use case specifically would be for embedding on a website

knadh:
There's a hack that you can try.

Paste the following (after tweaking) into Settings -> Appearance -> Public -> Javascript.

```javascript
(function() {
  let url;
  switch(new URLSearchParams(window.location.search).get('redirect')) {
      case 'first':
        url = 'https://google.com';
      break;
      case 'second:
        url = 'https://yahoo.com';
      break;
  }

  if (url) {
    window.location.href = url;
  }
})();
```

In the form code that's generated, append `?redirect=first` (or whatever name you apply in the script above). Eg:

```html
<form method="post" action="https://listmonk.yoursite.com/subscription/form?redirect=first" class="listmonk-form">
...
```

knadh:
[...]

PR #3092, opened 2026-06-17
feat: pre-fill email on subscription form via ?email= query param

PabloCastellano:


knadh:
Thanks @PabloCastellano, but don't think this is a good idea, giving an arbitrary injectible field on the public form. If you really need this behaviour, it can be hacked together with a few lines of JS targeting the form in Admin -> Settings -> Appearance -> Custom JS. Eg: https://github.com/knadh/listmonk/issues/2145#issuecomment-2661322635

PR #3096, opened 2026-06-18
Support `robots.txt` in static content

### LINKED ISSUE #3094: Add `robots.txt`
Listmonk returns 404 for `robots.txt`, and there's no way to provide one without hacking on Listmonk.

In these troubled times, crawlers crawl a _lot_ of content. Listmonk generates a lot of URLs with IDs in them, e.g. `/subscription/:campUUID/:subUUID`, `/link/:linkUUID/:campUUID/:subUUID`, `/campaign/:campUUID/:subUUID`. It hurts bandwidth to have crawlers spend their time on URLs like these.

Speaking for myself, I'd like to use a `robots.txt` like this:

```
User-agent: *
Allow: /subscription/form
Disallow: /subscription/
Disallow: /campaign/
Disallow: /link/
```

But, more importantly, I think we'd want `robots.txt` to be something that Listmonk users can customize via `--static-dir`, just like customizing the logo.

(It'd be fine if Listmonk shipped with a blank `robots.txt` out of the box, as long as I could provide my own.)

dfabulich:
Fixes #3094

### DIFF
--- a/cmd/handlers.go
+++ b/cmd/handlers.go
@@ -290,6 +290,7 @@ func initHTTPHandlers(e *echo.Echo, a *App) {
 
 		// Public health API endpoint.
 		g.GET("/health", a.HealthCheck)
+		g.GET("/robots.txt", a.RobotsTxt)
 
 		// 404 pages.
 		g.RouteNotFound("/*", func(c echo.Context) error {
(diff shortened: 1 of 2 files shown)


dfabulich:
I was a little uncertain where `robots.txt` should go in the `static/` directory. I settled on `static/public/static`, but I'd be happy to change it in response to feedback.

```
static/
  email-templates/
  public/
    static/
    templates/
```

`robots.txt` is static, and so I figured that it should go in `static/public/static`. All of the content there is automatically served up at `/public/static/*`, e.g. `/public/static/logo.png`. Putting it there makes `robots.txt` available at both `/robots.txt` _and_ at `/public/static/robots.txt`, which is kinda weird, but seems fine to me.

Alternately, we could put it in `static/public/templates`, so it fits in with the other documented "pages". But `robots.txt` isn't actually a "template"; we don't support injecting any of the Go template strings documented in https://listmonk.app/docs/templating/ in `robots.txt`.

PR #3102, opened 2026-06-19
`dev-docker` fixes

### LINKED ISSUE #3098: dev-docker make targets depend on `build`
The `init-dev-docker`, `dev-docker`, and even `rm-dev-docker` make targets all depend on `build`, which requires you to set up a working `go` build environment on the host machine.

None of these commands actually require the host to do a `go` build; the whole point of using Docker for the dev environment is so that I wouldn't have to get the `go` build set up.

### LINKED ISSUE #3099: `make init-dev-docker` fails due to old Node
`make init-dev-docker` in the current `master` branch. It will fail, with this error:

```
error marked@12.0.2: The engine "node" is incompatible with this module. Expected version ">= 18". Got "16.20.2"
```

<details><summary>Full log</summary>
<p>

```
dfab: ~/git/listmonk % make init-dev-docker
cd dev; \
	docker compose build ; \

WARN[0000] /Users/dfabulich/git/listmonk/dev/docker-compose.yml: the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion
[+] Building 0.6s (13/13) FINISHED
 => [internal] load local bake definitions                                                                                                                                                       0.0s
 => => reading from stdin 935B                                                                                                                                                                   0.0s
 => [backend internal] load build definition from app.Dockerfile                                                                                                                                 0.0s
 => => transferring dockerfile: 251B                                                                                                                                                             0.0s
 => [backend internal] load metadata for docker.io/library/node:16                                                                                                                               0.0s
 => [backend internal] load metadata for docker.io/library/golang:1.24.1                                                                                                                         0.4s
 => [front internal] load .dockerignore                                                                                                                                                          0.0s
 => => transferring context: 358B                                                                                                                                                                0.0s
 => [backend node 1/3] FROM docker.io/library/node:16                                                                                                                                            0.0s
 => [backend go 1/1] FROM docker.io/library/golang:1.24.1@sha256:52ff1b35ff8de185bf9fd26c70077190cd0bed1e9f16a2d498ce907e5c421268                                                                0.0s
 => CACHED [backend node 2/3] COPY --from=go /usr/local/go /usr/local/go                                                                                                                         0.0s
 => CACHED [backend node 3/3] WORKDIR /app                                                                                                                                                       0.0s
 => [backend] exporting to image                                                                                                                                                                 0.0s
 => => exporting layers                                                                                                                                                                          0.0s
 => => writing image sha256:b91ee87168c5aad8eb98f1bdb41b5be299024f3c3263c8d43fb5ae8bcfc03eb7                                                                                                     0.0s
 => => naming to docker.io/library/dev-backend                                                                                                                                                   0.0s
 => [front] exporting to image                                                                                                                                                                   0.0s
 => => exporting layers                                                                                                                                                                          0.0s
 => => writing image sha256:48d967343ddc3f8778b3455ca13e8521246c030f9168bea1e9c10ba6c4c655dc                                                                                                     0.0s
 => => naming to docker.io/library/dev-front                                                                                                                                                     0.0s
 => [front] resolving provenance for metadata file                                                                                                                                               0.0s
 => [backend] resolving provenance for metadata file                                                                                                                                             0.0s
[+] Building 2/2
 ✔ dev-front    Built                                                                                                                                                                            0.0s
 ✔ dev-backend  Built                                                                                                                                                                            0.0s
cd dev; \
	docker compose run --rm backend sh -c "make dist && ./listmonk --install --idempotent --yes --config dev/config.toml"
WARN[0000] /Users/dfabulich/git/listmonk/dev/docker-compose.yml: the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion
[+] Creating 3/3
 ✔ Network dev_listmonk-dev    Created                                                                                                                                                           0.0s
 ✔ Volume dev_listmonk-dev-db  Created                                                                                                                                                           0.0s
 ✔ Container dev-db-1          Created                                                                                                                                                           0.0s
[+] Running 1/1
 ✔ Container dev-db-1  Started                                                                                                                                                                   0.1s
go install github.com/knadh/stuffbin/...
CGO_ENABLED=0 go build -o listmonk -ldflags="-s -w -X 'main.buildString=nightly (#f4af2601 2026-06-19T20:29:29+0000)' -X 'main.versionString=nightly'" ./cmd
cd frontend/email-builder && yarn install
yarn install v1.22.19
[1/4] Resolving packages...
[2/4] Fetching packages...
error marked@12.0.2: The engine "node" is incompatible with this module. Expected version ">= 18". Got "16.20.2"
error Found incompatible module.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
make: *** [Makefile:57: frontend/email-builder/node_modules] Error 1

make: *** [init-dev-docker] Error 2
```

</p>
</details>

### LINKED ISSUE #3100: `make dev-docker` doesn't forward `http://localhost:8080` to the frontend service
(This might be specific to macOS, which runs Docker in a VM. I'm on macOS 26.5.1.)

```
make init-dev-docker
make dev-docker
```

Then `curl http://localhost:8080`; it'll fail, "curl: (56) Recv failure: Connection reset by peer".

Inside the frontend service, we're running `yarn dev` which runs `vite`, which binds to `http://localhost:8080`, but to forward the port from Docker, we need to run `vite --host` or `host: true` in Vite configuration, causing it to bind to `0.0.0.0:8080`, allowing Docker to forward the port to the host.

### LINKED ISSUE #3101: devcontainer fails to launch
```
git clean -xdf
make rm-dev-docker
```

Then open the project in Visual Studio Code, open command palette, and select "Dev Containers: Rebuild and Reopen in Container".

It will fail with an error "An error occurred setting up the container." If you view the Docker logs for `dev-backend-1`, you'll see this error:

```
2026-06-19 13:48:24.708 | Container started
2026-06-19 13:48:24.968 | CGO_ENABLED=0 go run -ldflags="-s -w -X 'main.buildString=nightly (#f4af2601 2026-06-19T20:48:24+0000)' -X 'main.versionString=nightly' -X 'main.frontendDir=frontend/dist'" ./cmd --config=dev/config.toml
2026-06-19 13:48:30.625 | 2026/06/19 20:48:30.625433 main.go:106: nightly (#f4af2601 2026-06-19T20:48:24+0000)
2026-06-19 13:48:30.625 | 2026/06/19 20:48:30.625483 init.go:191: reading config: dev/config.toml
2026-06-19 13:48:30.625 | 2026/06/19 20:48:30.625842 init.go:331: connecting to db: db:5432/listmonk-dev
2026-06-19 13:48:30.628 | 2026/06/19 20:48:30.628402 init.go:248: unable to initialize embedded filesystem (no ID found in the file). Using local filesystem
2026-06-19 13:48:30.628 | 2026/06/19 20:48:30.628414 init.go:270: loading i18n files from: i18n
2026-06-19 13:48:30.628 | 2026/06/19 20:48:30.628416 init.go:290: loading static files from: static
2026-06-19 13:48:30.631 | 2026/06/19 20:48:30.631172 init.go:302: failed reading static files from disk: 'static': stat frontend/dist: no such file or directory
2026-06-19 13:48:30.631 | exit status 1
2026-06-19 13:48:30.651 | make: *** [Makefile:137: run-backend-docker] Error 1
```

That is, it failed to launch because `make dist` hadn't run.

`devcontainer.json` says to run `make dist` in `postStartCommand`, but `postStartCommand` runs after the container has started, i.e. after its Docker Compose `command` (`make run-backend-docker`) has completed successfully. The `command` fails before `postStartCommand` can run.

### LINKED ISSUE #3104: Restarting the backend in the devcontainer restarts the devcontainer
The devcontainer is the `backend` container; its `command` is `make run-backend-docker`. If the backend dies for any reason, the whole devcontainer dies, forcing the VS Code to reconnect to the container.

Instead, the devcontainer should be its own container, a sibling of the `backend`.

### LINKED ISSUE #3111: `dev-docker` super issue
On #3098, @knadh wrote:

> There are 4 issues and 3 PRs related to this that are currently open. Could you please consolidate them?
>
> I don't use the dev-docker setup (it was a community contrib), so I've to study it. A consolidated issue+PR will help me make sense of this better.

I filed _five_ issues on dev-docker last week. They differ in scope; some of them could be fairly called "enhancements" to the `dev-docker` experience, whereas others are critical, blocking bugs.

Here, I'll provide background documentation on how the `dev-docker` environment works, and provide additional context on the issues I filed.

With this background, you should be in a good position to review my consolidated PR #3102 and merge it.

## Background

First, let's review https://listmonk.app/docs/developer-setup/

> You can run your dev environment locally or inside containers.
> 
> After setting up the dev environment, you can visit `http://localhost:8080`.
> 
> 
> 1. Locally
> 
>     - Run `make run` to start the listmonk dev server on `:9000`.
>     - Run `make run-frontend` to start the Vue frontend in dev mode using yarn on `:8080`. All `/api/*` calls are proxied to the app running on `:9000`. Refer to the [frontend README](https://github.com/knadh/listmonk/blob/master/frontend/README.md) for an overview on how the frontend is structured.
> 
> 2. Inside containers (Using Makefile)
> 
>     - Run `make init-dev-docker` to setup container for db.
>     - Run `make dev-docker` to setup docker container suite.
>     - Run `make rm-dev-docker` to clean up docker container suite.
> 
> 3. Inside containers (Using devcontainer)
> 
>     - Open repo in vscode, open command palette, and select "Dev Containers: Rebuild and Reopen in Container".
> 
> It will set up db, and start frontend/backend for you.

It's kinda weird that there are two different ways to build with containers, but it kinda makes sense, because the devcontainer mode 3 is partially built on top of the "Makefile" mode 2.

### Mode 2: Inside containers (Using Makefile)

Docker is mostly driven by the `dev` https://github.com/knadh/listmonk/blob/master/dev/docker-compose.yml file, which defines five services to launch:

  * `adminer`
  * `mailhog`
  * `db`
  * `backend`: runs `make run-backend-docker`, which runs `go run` to run the `listmonk` binary on port 9000
  * `frontend`: runs `make run-frontend`, which runs `yarn dev` on port 8080; that service proxies API requests to the backend on port 9000

The `backend` container requires one-time setup.
1. `backend` requires _some_ kinda local `frontend` build (the `frontend/dist` directory has to at least _exist_ with some content)
2. The `db` needs to be installed

`make init-dev-docker` boots up the `backend` container and runs `make dist && ./listmonk --install --idempotent --yes --config dev/config.toml`, fulfilling both parts of the one-time setup.

(It's kinda unfortunate that `make init-dev-docker` runs `make dist`, a full production dist build including minification, instead of `make frontend-build`. We'll get back to that later on.)

Once that's done, `make dev-docker` will run `docker compose up` in the `dev` directory, launching all of the services synchronously. Their logs appear interleaved on the terminal; you can stop all of the dev containers at once with Ctrl-C.

To clear out their state, you can run `make rm-dev-docker`, which runs `docker compose down -v` which means to stop the containers (they were probably already stopped) _and_ `-v` deletes their "volumes," deleting DB state.

### Mode 3: Inside containers (Using devcontainer)

Devcontainer mode is defined by https://github.com/knadh/listmonk/blob/master/.devcontainer/devcontainer.json

It's designed to automate launching the `dev` https://github.com/knadh/listmonk/blob/master/dev/docker-compose.yml file in a standard way that works in one-click, instead of manual instructions.

If you open listmonk in VS Code, the first thing it will do is pop up a notification saying: "Folder contains a Dev Container configuration file. Reopen folder to develop in a container" with a big old "Reopen in container" button. If you click that button, VS Code is supposed to automatically launch `docker compose`, automatically do one-time initialization, and then (this is kinda weird), shut down the currently running window and run a copy of VS Code _inside_ one of the containers.

Even better, Github has this nifty feature where you can navigate to https://github.com/knadh/listmonk/ click the green "Code" button and select the "Codespaces" tab. If you do that, it's supposed to automatically set up a working dev environment in your devcontainer inside VS Code, right there in your browser. No downloading, no setup, dev ready to go.

Unfortunately, the devcontainer mode doesn't work in the current `master` branch, as we'll see.

### #3098
* #3098
  > The `init-dev-docker`, `dev-docker`, and even `rm-dev-docker` make targets all depend on `build`, which requires you to set up a working `go` build environment on the host machine.
  > 
  > None of these commands actually require the host to do a `go` build; the whole point of using Docker for the dev environment is so that I wouldn't have to get the `go` build set up.
  
I would describe this as a "blocker with a workaround." The workaround is to build listmonk once locally, which requires installing a compatible version of Go, etc. But, as I pointed out, the whole point of using Docker for the dev environment is so that I wouldn't have to do that.

The fix for this is very brief. In https://github.com/knadh/listmonk/pull/3102/changes/c2e33e54c5592bc753de3f850805bd26aafa640b, I just removed `build` as a dependency from a couple of `Makefile` targets. They didn't _actually_ depend on the `listmonk` binary being created locally, so this dependency was totally unnecessary.

### #3099
* #3099
  > `make init-dev-docker` in the current `master` branch. It will fail, with this error:
  > 
  > ```
  > error marked@12.0.2: The engine "node" is incompatible with this module. Expected version ">= 18". Got "16.20.2"
  > ```

This one's an absolute blocker. (I have to assume that nobody at all has been using the `dev-docker` workflow for quite some time…?)

The beauty of Docker is that you can pin your dependencies in a `Dockerfile`. The downside is that somebody has to remember to keep those dependencies up to date.

Fixing this is a one-line change https://github.com/knadh/listmonk/pull/3102/changes/6b61486e9ff5d322c1f6a6b47339563f41cc0df1 changing the pinned Node `16` to the latest LTS Node `24`.

### #3100 
* #3100

> (This might be specific to macOS, which runs Docker in a VM. I'm on macOS 26.5.1.)
> 
> ```
> make init-dev-docker
> make dev-docker
> ```
> 
> Then `curl http://localhost:8080`; it'll fail, "curl: (56) Recv failure: Connection reset by peer".
> 
> Inside the frontend service, we're running `yarn dev` which runs `vite`, which binds to `http://localhost:8080`, but to forward the port from Docker, we need to run `vite --host` or `host: true` in Vite configuration, causing it to bind to `0.0.0.0:8080`, allowing Docker to forward the port to the host.

This issue blocks development in Mode 2. (This again makes me wonder if anybody's ever been using this thing… 🤔)

Hopefully you're already familiar with the difference between localhost and 0.0.0.0, but, just in case…

<details><summary>The difference between localhost and 0.0.0.0</summary>
<p>

When you bind a TCP port, you get to decide whether external hosts can connect to the port, or whether the port is only accessible from the localhost current machine. If you bind `localhost:8080`, you can `curl http://localhost:8080`, but people on other machines can't reach that port, even if they're on the same wifi network and they know your IP address.

If you bind 0.0.0.0:8080, anyone who knows your IP address can contact you at that port. `curl http://localhost:8080` will work either way, but other machines on the same network can `curl http://192.168.0.17:8080` (or whatever your IP address is) to talk to you.

It's a good security practice to avoid binding 0.0.0.0 unless you actually need to run a server.

</p>
</details> 

There is a bit of a subtlety around how Docker works on macOS vs. Linux. On Linux, Docker containers are lighter weight than virtual machines, which run their own (emulated?) CPU, reserve a fixed amount of RAM, and claim their own IP address. On macOS, Docker containers are actually VMs; they're literally running QEMU.

As a result, `localhost` can mean different things to Docker containers on Linux vs. Docker containers on macOS. On macOS, the Docker container is always a "different machine." It has its own IP address, so if you want to talk to services on port 8080, they need to serve themselves up to "external" hosts (i.e. the host machine running the VM).

On Linux, containers _can_ have their own IP address, but usually, by default, they're running on the "same machine." Depending on how you configure Docker, binding `localhost:8080` might well let you access the container's service from the host, since the host is the same localhost. (I'm a bit hazy on how this is configured on Linux.)

What I do know for sure is that on macOS, in Mode 2, the first thing you need to do is to contact `http://localhost:8080`, and it's not going to work, because that port is only bound inside the container's VM.

The fix could have been as simple as just setting `host: true` in https://github.com/knadh/listmonk/blob/master/frontend/vite.config.js but that's a security-sensitive thing to do. If you're building Listmonk locally, without Docker, you probably don't want `host: true` there. So, instead, I set up `docker-compose.yml` to pass an environment variable to `make`. In `vite.config.js` we check to see whether `env.LISTMONK_VITE_HOST === 'true'`, and set `host: true` only in that case.

With `host: true` enabled, Mode 2 "Inside containers (using Makefile)" works.

(My initial draft of PR #3102 only fixed Mode 2, and only fixed #3098, #3099, and #3100. I filed separate PRs to fix Mode 3.)

### #3101
* #3101

This issue 3101 only impacts devcontainer mode. It's a "blocker" for the devcontainer mode, but the obvious workaround is to just switch to mode 2, "Inside containers (using Makefile)."

But 3101 is pretty obnoxious for new developers, because VS Code offers a prominent "Reopen in container" button; it looks like you should probably click it. If you do, VS Code will try to relaunch itself inside the container, and then fail to function at all as an editor when you hit this bug.

And, of course, this bug totally breaks Github Codespaces; you can't hack on listmonk in a browser right now as a result.

With that out of the way, here's the issue:

> ```
> git clean -xdf
> make rm-dev-docker
> ```
> 
> 
>     
>   
> 
> Then open the project in Visual Studio Code, open command palette, and select "Dev Containers: Rebuild and Reopen in Container".
> 
> 
> It will fail with an error "An error occurred setting up the container." If you view the Docker logs for `dev-backend-1`, you'll see this error:
> 
> 
> ```
> 2026-06-19 13:48:24.708 | Container started
> 2026-06-19 13:48:24.968 | CGO_ENABLED=0 go run -ldflags="-s -w -X 'main.buildString=nightly (#f4af2601 2026-06-19T20:48:24+0000)' -X 'main.versionString=nightly' -X 'main.frontendDir=frontend/dist'" ./cmd --config=dev/config.toml
> 2026-06-19 13:48:30.625 | 2026/06/19 20:48:30.625433 main.go:106: nightly (#f4af2601 2026-06-19T20:48:24+0000)
> 2026-06-19 13:48:30.625 | 2026/06/19 20:48:30.625483 init.go:191: reading config: dev/config.toml
> 2026-06-19 13:48:30.625 | 2026/06/19 20:48:30.625842 init.go:331: connecting to db: db:5432/listmonk-dev
> 2026-06-19 13:48:30.628 | 2026/06/19 20:48:30.628402 init.go:248: unable to initialize embedded filesystem (no ID found in the file). Using local filesystem
> 2026-06-19 13:48:30.628 | 2026/06/19 20:48:30.628414 init.go:270: loading i18n files from: i18n
> 2026-06-19 13:48:30.628 | 2026/06/19 20:48:30.628416 init.go:290: loading static files from: static
> 2026-06-19 13:48:30.631 | 2026/06/19 20:48:30.631172 init.go:302: failed reading static files from disk: 'static': stat frontend/dist: no such file or directory
> 2026-06-19 13:48:30.631 | exit status 1
> 2026-06-19 13:48:30.651 | make: *** [Makefile:137: run-backend-docker] Error 1
> ```
> 
> 
>     
> 
>   
> 
> That is, it failed to launch because `make dist` hadn't run.
> 
> 
> `devcontainer.json` says to run `make dist` in `postStartCommand`, but `postStartCommand` runs after the container has started, i.e. after its Docker Compose `command` (`make run-backend-docker`) has completed successfully. The `command` fails before `postStartCommand` can run.

Remember earlier when I mentioned that the `backend` container requires one-time setup? In Makefile mode, we ask developers to run `make init-dev-docker` first, but devcontainers don't support that. The `postStartCommand` approach seems like it would do what we want, but AFAIK it never worked, because it would only run the commands if you'd already ran `make init-dev-docker`, i.e. `postStartCommand` only works if you don't need it to work!

The fix for this one is a little bit more bold than my other commits, which is why I'd initially filed a separate PR for it. https://github.com/knadh/listmonk/pull/3102/changes/be5c82a1c0a615702aaac8cb6960062787e42fe3

> **Initialize `dev-docker` automatically**
> 
> `init-dev-docker` ran `make dist && ./listmonk --install`, and was an explicit manual step that users would have to do to get the `dev-docker` environment up and running.
> 
> For users of the `dev-docker` development environment, this commit simplifies setup; users just have to `make dev-docker` without having to remember to `make init-dev-docker`.
> 
> But, more importantly, this fixes the broken `devcontainer.json`, which gave users no opportunity to `init-dev-docker` before it automatically tries to start the `backend` container.
> 
> Now, `backend` depends on a new `init` container, which runs `make init-docker` and exits. `make init-docker` runs `make build-frontend` (we never needed a production `make dist` build for development), builds `listmonk`, and runs `./listmonk --install --idempotent`.
> 
> On my machine, this imperceptibly slows down `dev-docker` startup, running `./listmonk --install --idempotent` before starting the `backend`.

This felt like a big change, not just to devcontainer mode, but to the Makefile `dev-docker` mode. It completely eliminates the need to run `make init-dev-docker`, but it does it in a way that causes even the Makefile `dev-docker` mode to _always_ run `./listmonk --install --idempotent` before launching the `backend` server, causing (for me) an imperceptible slowdown every time you launch the backend.

Also, I took this opportunity to change the initialization step. `make init-dev-docker` ran `make dist`, which is much slower than `make frontend-build`, and totally unnecessary, because we're probably never going to use any of the frontend code that the `backend` code depends on. It exists there only so that the `backend` _has_ some static content at all. Switching it to run `make frontend-build` made that initial setup much faster on my machine.

Is all that OK? I hope it is. It definitely fixed devcontainer mode on my machine.

### #3104
* #3104

This one is, debatably, just an enhancement to the devcontainer workflow, but IMO it's a pretty big enhancement.

As I mentioned before, when VS Code runs a devcontainer, it doesn't just automatically setup and start the containers. It closes the current window and spawns a new window running inside the devcontainer.

> The devcontainer is the `backend` container; its `command` is `make run-backend-docker`. If the backend dies for any reason, the whole devcontainer dies, forcing the VS Code to reconnect to the container.
> 
> Instead, the devcontainer should be its own container, a sibling of the `backend`.

So, after all of my other commits have applied, you can finally open Listmonk in VS Code. The backend and frontend automatically set themselves up. localhost:8080 works. You make a small change to the Go code. … now what?

If you're using Mode 2 (Makefile), you ran the containers using `make dev-docker`; Ctrl-C stops them. Then, you can `make dev-docker` again to rebuild the backend and test your changes.

If you're using Mode 3, you need to stop the backend to restart it. But, if you stop the backend, you'll stop the _devcontainer_, the container in which your _IDE_ is running.

In other words, if you stop the backend, your text editor stops working! 😬 

The fix is mildly more complicated, and adds a _dependency_, so we should examine this with extra scrutiny. https://github.com/knadh/listmonk/pull/3102/changes/2f83c87e385058b53b9b0c23e99005b417d2d770

What we do is add a new service to `dev/docker-compose.yml` called `devcontainer`. The `devcontainer` service does "nothing." Its `command` is `sleep infinity`. Its sole purpose is to let VS Code connect to it and live there.

Once that's set up, you still need some way to actually restart the backend container. You can do it from the host, with `cd dev; docker compose restart backend`, but that's kinda annoying, especially if you were running an LLM inside VS Code, because it means that the LLM can't restart the backend itself.

So, to make this workable, I wanted to add Docker commands inside the `devcontainer`. It turns out there's a "[Feature](https://github.com/devcontainers/features/pkgs/container/features)" for that. "Feature" is the devcontainer term for plugins, basically.

> 'Features' are self-contained units of installation code and development container configuration. Features are designed to install atop a wide-range of base container images.
> 
> Missing a CLI or language in your otherwise perfect container image? Add the relevant Feature to the features property of a [devcontainer.json](https://containers.dev/implementors/json_reference/#general-properties). A [tool supporting the dev container specification](https://containers.dev/supporting) is required to build a development container.
> 
> You may learn about Features at [containers.dev](https://containers.dev/implementors/features/), which is the website for the dev container specification.

The <s>plugin</s> Feature I'm using is an official devcontainer Feature called "[Docker outside of Docker](https://github.com/devcontainers/features/tree/main/src/docker-outside-of-docker)," abreviated "DooD." The readme for DooD says:

> Re-use the host docker socket, adding the Docker CLI to a container. Feature invokes a script to enable using a forwarded Docker socket within a container to run Docker commands.

I think that's a very confusing way to say what DooD does. DooD allows you to run Docker commands inside a Docker container, allowing a Docker container to read logs on its sibling containers, and to start/stop other containers.

The peril of DooD is that it's not restricted to Docker containers in the same group. Using DooD, you can read logs on _any_ Docker container on the current machine, not just listmonk containers. I think that's OK for a devcontainer, but someone might disagree.

With DooD setup, you can `docker compose restart backend` inside the devcontainer, and it just works. The IDE remains connected to the `devcontainer` container, and everything works nicely.

DooD _is_ kinda creepy though. So I also updated the Mode 2 Makefile targets to explicitly list out the services they're supposed to start, intentionally excluding `devcontainer`. That way, Mode 2 Makefile developers won't start a DooD devcontainer that they won't even use.

(Some day, it'd be nice to install https://github.com/air-verse/air for hot reloading the Go code. But _that's_ an enhancement for another day!)

dfabulich:
* Fixes #3111
  * Fixes #3098
  * Fixes #3099
  * Fixes #3100
  * Fixes #3101
  * Fixes #3104

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -3,6 +3,5 @@
   "dockerComposeFile": "../dev/docker-compose.yml",
   "service": "backend",
   "workspaceFolder": "/app",
-  "forwardPorts": [9000],
-  "postStartCommand": "make dist && ./listmonk --install --idempotent --yes --config dev/config.toml"
+  "forwardPorts": [9000]
 }
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,7 +1,10 @@
 {
   "name": "listmonk",
   "dockerComposeFile": "../dev/docker-compose.yml",
-  "service": "backend",
+  "service": "devcontainer",
   "workspaceFolder": "/app",
-  "forwardPorts": [9000]
+  "runServices": ["adminer", "mailhog", "db", "backend", "front"],
+  "features": {
+    "ghcr.io/devcontainers/features/docker-outside-of-docker:1": {}
+  }
 }
--- a/Makefile
+++ b/Makefile
@@ -121,7 +121,7 @@ release:
 
 # Build local docker images for development.
 .PHONY: build-dev-docker
-build-dev-docker: build ## Build docker containers for the entire suite (Front/Core/PG).
+build-dev-docker: ## Build docker containers for the entire suite (Front/Core/PG).
 	cd dev; \
 	docker compose build ; \
 
@@ -138,7 +138,7 @@ run-backend-docker:
 
 # Tear down the complete local development docker suite.
 .PHONY: rm-dev-docker
-rm-dev-docker: build ## Delete the docker containers including DB volumes.
+rm-dev-docker: ## Delete the docker containers including DB volumes.
 	cd dev; \
 	docker compose down -v ; \
 
--- a/Makefile
+++ b/Makefile
@@ -136,6 +136,13 @@ dev-docker: build-dev-docker ## Build and spawns docker containers for the entir
 run-backend-docker:
 	CGO_ENABLED=0 go run -ldflags="-s -w -X 'main.buildString=${BUILDSTR}' -X 'main.versionString=${VERSION}' -X 'main.frontendDir=frontend/dist'" ./cmd --config=dev/config.toml
 
+# Build assets and install the DB schema for the docker dev suite.
+.PHONY: init-docker
+init-docker:
+	@test -d frontend/dist || $(MAKE) build-frontend
+	@test -f $(BIN) || $(MAKE) build
+	./listmonk --install --idempotent --yes --config dev/config.toml
+
 # Tear down the complete local development docker suite.
 .PHONY: rm-dev-docker
 rm-dev-docker: ## Delete the docker containers including DB volumes.
@@ -144,6 +151,6 @@ rm-dev-docker: ## Delete the docker containers including DB volumes.
 
 # Setup the db for local dev docker suite.
 .PHONY: init-dev-docker
-init-dev-docker: build-dev-docker ## Delete the docker containers including DB volumes.
+init-dev-docker: build-dev-docker ## Build assets and install the DB schema in docker.
 	cd dev; \
-	docker compose run --rm backend sh -c "make dist && ./listmonk --install --idempotent --yes --config dev/config.toml"
+	docker compose run --rm init
--- a/Makefile
+++ b/Makefile
@@ -119,6 +119,9 @@ release-dry:
 release:
 	goreleaser release --parallelism 1 --clean
 
+# App services for `make dev-docker` (excludes the devcontainer workspace).
+DEV_DOCKER_SERVICES = adminer mailhog db backend front
+
 # Build local docker images for development.
 .PHONY: build-dev-docker
 build-dev-docker: ## Build docker containers for the entire suite (Front/Core/PG).
@@ -129,7 +132,7 @@ build-dev-docker: ## Build docker containers for the entire suite (Front/Core/PG
 .PHONY: dev-docker
 dev-docker: build-dev-docker ## Build and spawns docker containers for the entire suite (Front/Core/PG).
 	cd dev; \
-	docker compose up
+	docker compose up $(DEV_DOCKER_SERVICES)
 
 # Run the backend in docker-dev mode. The frontend assets in dev mode are loaded from disk from frontend/dist.
 .PHONY: run-backend-docker
--- a/dev/app.Dockerfile
+++ b/dev/app.Dockerfile
@@ -1,6 +1,6 @@
 FROM golang:1.24.1 AS go
 
-FROM node:16 AS node
+FROM node:24 AS node
 
 COPY --from=go /usr/local/go /usr/local/go
 ENV GOPATH /go
--- a/dev/docker-compose.yml
+++ b/dev/docker-compose.yml
@@ -42,6 +42,7 @@ services:
       - "8080:8080"
     environment:
       - LISTMONK_API_URL=http://backend:9000
+      - LISTMONK_VITE_HOST=true
     depends_on:
       - db
     volumes:
--- a/dev/docker-compose.yml
+++ b/dev/docker-compose.yml
@@ -83,6 +83,24 @@ services:
     networks:
       - listmonk-dev
 
+  # Workspace container for VS Code / Cursor devcontainers. Omitted from
+  # `make dev-docker`; the editor starts this service when reopening in container.
+  devcontainer:
+    image: dev-devcontainer
+    build:
+      context: ../
+      dockerfile: dev/app.Dockerfile
+    command: ["sleep", "infinity"]
+    volumes:
+      - ../:/app
+      - ${GOPATH:-${HOME}/go}/pkg/mod/cache:/go/pkg/mod/cache
+      - /var/run/docker.sock:/var/run/docker.sock
+    depends_on:
+      db:
+        condition: service_started
+    networks:
+      - listmonk-dev
+
 volumes:
   listmonk-dev-db:
 
--- a/dev/docker-compose.yml
+++ b/dev/docker-compose.yml
@@ -33,6 +33,21 @@ services:
         source: listmonk-dev-db
         target: /var/lib/postgresql/data
 
+  init:
+    build:
+      context: ../
+      dockerfile: dev/app.Dockerfile
+    command: ["make", "init-docker"]
+    depends_on:
+      db:
+        condition: service_started
+    volumes:
+      - ../:/app
+      - ${GOPATH:-${HOME}/go}/pkg/mod/cache:/go/pkg/mod/cache
+    networks:
+      - listmonk-dev
+    restart: "no"
+
   front:
     build:
       context: ../
@@ -58,7 +73,10 @@ services:
     ports:
       - "9000:9000"
     depends_on:
-      - db
+      db:
+        condition: service_started
+      init:
+        condition: service_completed_successfully
     volumes:
       - ../:/app
       - ${GOPATH:-${HOME}/go}/pkg/mod/cache:/go/pkg/mod/cache
--- a/docs/docs/content/developer-setup.md
+++ b/docs/docs/content/developer-setup.md
@@ -30,8 +30,7 @@ After setting up the dev environment, you can visit `http://localhost:8080`.
 
 2. Inside containers (Using Makefile)
 
-    - Run `make init-dev-docker` to setup container for db.
-    - Run `make dev-docker` to setup docker container suite.
+    - Run `make dev-docker` to build and start the docker container suite.
     - Run `make rm-dev-docker` to clean up docker container suite.
 
 3. Inside containers (Using devcontainer)
--- a/frontend/.env.sample
+++ b/frontend/.env.sample
@@ -1,2 +1,4 @@
 LISTMONK_FRONTEND_PORT=8080
 LISTMONK_API_URL="http://127.0.0.1:9000"
+# Bind Vite to 0.0.0.0 (required for Docker port publishing). Set by dev/docker-compose.yml.
+# LISTMONK_VITE_HOST=true
--- a/frontend/vite.config.js
+++ b/frontend/vite.config.js
@@ -21,6 +21,7 @@ export default defineConfig(({ _, mode }) => {
     },
     server: {
       port: env.LISTMONK_FRONTEND_PORT || 8080,
+      host: env.LISTMONK_VITE_HOST === 'true',
       proxy: {
         '^/$': {
           target: env.LISTMONK_API_URL || 'http://127.0.0.1:9000',


dfabulich:
> [P1] Rebuild listmonk inside init-docker instead of skipping on file exist

I disagree with characterizing this as critical. The `init` step is supposed to be one-time initialization. Running `make build`, checking timestamps on N source files, running a build, just so we can run `./listmonk --install --idempotent` seems like a bad move on a very hot path for dev reloading.

If we _really_ want to turn this into a `make build` step, I'll do it, but in that case, I think we should prioritize adding https://github.com/air-verse/air live reloading to the dev environment, ensuring that developers don't have to restart the Go server (Incurring this cost) every time they want to make a small change to a Go file.

@knadh I don't want to make this change based on an LLM's suggestion; LMK if you agree with hodor here, in which case, I'll change it.

(BTW, this sort of thing is part of the reason why I had made these commits in a separate PR; I knew that the first three commits were _very_ safe, but the other two commits were more controversial.)

knadh:
Hi @dfabulich. I'll come back to this after v7 is ready (which fully eliminates the Vue frontend build).

Issue #2221, opened 2024-12-26
Determining which emails in campaign failed to send


arielelkin:
I sent out a campaign, "Service not available" error messages appeared and then got this final result:

<img width="420" alt="image" src="https://github.com/user-attachments/assets/447ffd47-33c1-4c3a-8461-53051ca5e468" />

How do I know which subscribers didn't receive the campaign? The logs don't indicate that.

MaximilianKohler:
Go to the list of subscribers, click on the last page. The last subscriber there gets emailed first, and so on.

Listmonk emails subscribers in reverse order in the list, so you'll have to count and compare with how many subscribers in the list vs how many emails were sent.

arielelkin:
I’ve been digging into this, but even with those lists, it’s still unclear who actually got the email and who didn’t.

This is the second time this has happened, and it’s really frustrating. Not knowing which subscribers missed the campaign makes it impossible to fix or follow up properly. If a campaign fails to send, the system should handle it in a way that helps users recover, not leave them in the dark.

A campaign marked as "completed" absolutely needs to show who got the email and who didn’t. Otherwise, how are we supposed to trust the system or manage our campaigns effectively? This feels like a basic feature that’s missing.

Kamalito75:
I have the same problem
![Image](https://github.com/user-attachments/assets/3f145f7e-28ed-427e-8998-55cf95a70d22)

Nordlicht-13:
I send my first two campaigns, and I got 250/251 and 292/293 send emails.
Why i'm always missing one? How do I know wich email was not send?

MaximilianKohler:
In my case it's usually due to a blocklisted contact. I don't think listmonk has a way to figure out which email it was yet. But if you use Amazon SES you can use https://sesdashboard.com.

arielelkin:
Bumping to prevent automated closing.

alex-simm:
As a solution, how about adding a button to finished campaigns that creates a new subscriber list consisting of all addresses which failed. Then the campaign (or a copy of it) can be sent again to that new list, making sure that everyone receives the email.

Issue #2079, opened 2024-10-10
Campaign: continue / resume / retry failed


wolfgang-braun:
Hi

due to exceeded quota not all e-mails have been sent after finishing campaign.
<img width="375" alt="Campaign" src="https://github.com/user-attachments/assets/39a2dd22-be68-4b46-85dd-447d514da442">
All clear so far.

Is it possible to send the email to the remaining participants? 
I can't find a UI element to trigger this. Maybe queries can be used to create a custom list?

The list should contain 520 (1902-1382) recipients .

Thank you!

MaximilianKohler:
You can try to resend based on views using:

Go to Subscribers list, click on "Advanced" button to use SQL query, and put query like this:
```
EXISTS(SELECT 1 FROM campaign_views WHERE campaign_views.subscriber_id=subscribers.id AND campaign_views.campaign_id=<put_id_of_campaign>)
```

How someone resent emails to subscribers that didn't get them: https://github.com/knadh/listmonk/issues/1629#issuecomment-1858762315 - but if you don't have anything in your logs that won't work.

wolfgang-braun:
@MaximilianKohler 
Thank you! This is an interesting approach. 
I wondered whether there's a table that tracks the actual email sending (attempts).
Seemingly not:
<img width="192" alt="table" src="https://github.com/user-attachments/assets/db265f09-c040-4b9a-a5f0-25b2a810dac2">

What we have is:
<img width="340" alt="track" src="https://github.com/user-attachments/assets/647eb4ed-3b3a-4352-9414-2d774ddb3a8f">
I think I can work with this to create a list of the missing 520 subscribers.

Any idea what's the ratio behind not tracking the sendings in a dedicated table?

mevinbabuc:
I tried something like this, to get the subscriber_id where the campaign stopped sending.
the numbers are not perfectly tallying (3 records here and there), but they are close enough.
@knadh can comment if this is the right way to get people

```WITH RankedRows AS (
    SELECT 
        subscriber_id, 
        ROW_NUMBER() OVER (ORDER BY subscriber_id) AS row_num
    FROM subscriber_lists where list_id = 30
)
SELECT subscriber_id
FROM RankedRows
WHERE row_num = 10163;
```

knadh:
Hi @mevinbabuc. Have you checked the `last_subscriber_id` in the `campaigns` table? If that doesn't match, then what you've done should work.

To handle this internally in listmonk is on the TODO.

[...]

Issue #3073, opened 2026-06-03
v7.0.0 + a note on new PRs


knadh:
**IMPORTANT** June 2026: Until the next major version v7.0.0 is released, it would not be possible to accept code contributions (PRs) that involve changes to the admin frontend/UI. The entire Vue/Buefy frontend is being removed completely in favour of a server-side-rendered admin UI that eliminates the Javascript-ecosystem lockin. listmonk admin UI will no longer be a Vue SPA (Single-page App) after v7.

-----------

6+ years ago, I was forced to spend close to 100 hours on completely rewriting the entire listmonk frontend from React+AntDesign UI to Vue 2+Buefy UI (https://github.com/knadh/listmonk/commit/97583fe4b415d73b80ba8a5ee1b101d423e66657). That was the price I had to pay for making the mistake of choosing React originally--a terrible, clunky, overengineered, bloated piece of technology that has contributed significantly to the bloating and enshittification of the WWW in general.

Fast forward a few years, Vue 2 has moved on to Vue 3. Buefy (which is built on top of the CSS library Bulma) has obsoleted the version listmonk uses, and has moved on to a hard dependency on Vue 3, leaving listmonk stranded.

* What listmonk has: Vue 2 app -> Buefy pre-v1 -> Bulma 0.xxx
* What Buefy v1 depends on: Vue 3 -> Bulma v1+
* Want improvements in Buefy? Must move the entire app to Vue 3.
* Want improvements in Bulma v1? Must move to the new Buefy and Vue 3.
* Can't upgrade to Vue 3? Stuck on old Buefy + old Bulma forever.

I wasted several hours trying to upgrade/rewire these dependencies, and it broke way too many things. It also surfaced [terrible technical choices](https://github.com/jgthms/bulma/issues/3811) in the Bulma library. I couldn't make it work despite LLM assistance (!). Yet again, the hot mess and dependency hell that is the Javascript "ecosystem" is a curse that [keeps on giving](https://nadh.in/blog/javascript-ecosystem-software-development-are-a-hot-mess/).

After reaching my wits' end, a couple of months ago, I made [Oat](https://oat.ink), an ~8KB, zero-dependency, semantic JS+CSS UI library. I've been using it in many of my projects and have gained significant confidence in it. I'm personally done with framework-loaded SPAs (Single Page Apps) for good. Very few apps out there have the legitimate need of being an SPA.

[...]

PR #3116, opened 2026-06-25
Resend failed subscribers

fkaiserbio:
This addresses a critical feature we were find missing. The ability to get the subscribers to which an email was not send.

We were having issues with our installation that did not allow to track down individual subscribers who did not receive a campaign. Especially when pausing and resuming a campaign, some issues occurred, which did fully send the campaign to all subscribers in the list.

The feature addresses the issues  #2221  #2079 which are in our opinion absolutely essential.

With a new button next in the campaigns overview you can now extract all subscribers, who did not receive the respective campaign into a new subscribers list. This allows to resend a campaign only to the subscribers which were not addressed for whatever error and eliminates the need to resend to everyone.

For making this work it was necessary to add the column `campaigns_sent`  to the `subscribers` table. It records every successfully sent campaign identified by its `campaign_id` and a timestamp.

The implementation was vibe coded. If you find it useful, you might consider a merge. We are using it in our installation.

Screenshot_20260625_122833


knadh:
Hi @fkaiserbio. Using a JSON field for logging isn't ideal.
- No foreign key support in the `campaigns_set` field. Campaigns can get deleted, but the log blob will remain.
- Without foreign keys, maintenance/cleanup of JSON data across millions of subscribers is going to be difficult.
- Every resend, even if it's to a dozen subscribers, requires scanning the entire subscribers table (which could have millions of records).
- The current approach mutates the `subscriber_lists` table, which it shouldn't.
- `Resending` failed messages has been a long pending request. That's what I'd like to focus on first, rather than resending a campaign to newly added users. Two different functions entirely.

Also, this PR involves changes to the obsolete Vue frontend. Please see: https://github.com/knadh/listmonk/issues/3073

I'll close this PR for now and would like to tackle these two features separately after v7. Thank you.

PR #3120, opened 2026-06-28
feat: add tag filter to campaigns list UI

### LINKED ISSUE #3119: feat: add tag filter to campaigns list UI
**Is your feature request related to a problem? Please describe.**
The campaigns list has no way to filter by tag. Users who manage multiple projects or clients via tags have no way to narrow the campaign list — they can only use the free-text search, which doesn't match on tags.

Add a tag filter input/dropdown next to the existing search bar on the campaigns list page. The backend already supports filtering by tag via `?tag=foo` (added in #651) — this is purely a frontend change to expose that capability.

The tag list can be populated from the tags already present on campaign objects (no new API endpoint needed).

vianmora:
## Summary

The campaigns API already supports filtering by tag (`?tag=`) but the
campaigns list page had no UI to use it.

- Adds a tag autocomplete input next to the existing search bar
- Tags are suggested from campaigns already loaded (accumulated across
  pages, no extra API call)
- Selecting/removing a tag immediately refreshes the list and resets to
  page 1
- Multiple tags can be combined

Closes #3119

## Test plan

- [ ] Add tags to a few campaigns
- [ ] Open the campaigns list, type a tag name → suggestions appear
- [ ] Select a tag → list filters correctly
- [ ] Add a second tag → list narrows further
- [ ] Remove a tag → list updates
- [ ] Pagination resets to page 1 when a tag is added/removed

### DIFF
--- a/frontend/src/views/Campaigns.vue
+++ b/frontend/src/views/Campaigns.vue
@@ -23,7 +23,7 @@
       :per-page="campaigns.perPage" :total="campaigns.total" hoverable checkable backend-sorting @sort="onSort">
       <template #top-left>
         <div class="columns">
-          <div class="column is-6">
+          <div class="column is-5">
             <form @submit.prevent="getCampaigns">
               <div>
                 <b-field>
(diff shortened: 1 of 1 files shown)


knadh:
Thanks. I'm adding this in the SSR v7 port. About PRs for the admin UI, please see: https://github.com/knadh/listmonk?tab=contributing-ov-file

PR #3127, opened 2026-07-01
Add optional per-campaign subscriber query for send-time segmentation

### LINKED ISSUE #250: Send campaign to segment of list

CookedApps:
Closes #250.

Adds an optional `subscriber_query` on a campaign: a WHERE expression (the same advanced query the Subscribers view already uses) that gets AND-ed onto the existing list/opt-in filter at send time. This lets you send a campaign to a subset of its lists' subscribers without the throwaway-list workaround.

This is the small version of #250: no saved segments, no dynamic lists (that was #2546, which you said was too big to review). One nullable column and a send-time filter.

### Why not just use a temp list

The usual workaround is query subscribers, copy them to a new list, send, delete the list. That breaks consent: unsubscribe links point at the temp list, one-click unsubscribe (RFC 8058) dies when the list is deleted, and the segment is stale the moment an attribute changes. Keeping the filter on the campaign and running it against the real lists means unsubscribe and opt-in stay bound to the actual lists. The filter can only ever remove recipients, never add them.

### What's in it

- Nullable `campaigns.subscriber_query TEXT`. Null/empty means whole list, exactly as today.
- At send time it's AND-ed onto the existing consent predicate. Blocklist, unsubscribes and double opt-in are untouched.
- Regular campaigns only. Validated on save, gated by the existing `subscribers:sql_query` permission (no new permission).
- Optional `POST /api/campaigns/preview-recipients` for an on-demand recipient count. Never as-you-type, since counting on a large instance is expensive.

### How it reuses what's there

- Same `%query%` splice and `validateQueryTables()` as the subscriber bulk query: `EXPLAIN (FORMAT JSON)` in a read-only tx, rejecting anything outside the existing table allowlist. Validated once on save; only the stored string is spliced at send.
- One new query, `next-campaign-subscribers-filtered`, is `next-campaign-subscribers` with `AND %query%` added. Same consent CASE, same `SELECT DISTINCT` so multi-list campaigns don't double-send. Campaigns without a query hit the unchanged prepared statement, so there's no regression for anyone not using this.
- `to_send` is recomputed once when the campaign's pipe starts, since the batched `next-campaigns` query can't splice a per-campaign predicate.

### Migration

`v6.3.0`: `ALTER TABLE campaigns ADD COLUMN IF NOT EXISTS subscriber_query TEXT NULL`. Additive, instant on large tables, nothing else changes.

### Frontend

The PR includes a small campaign-form editor (segment field plus a preview button) in one isolated commit. I know the Vue admin is frozen until v7, so this part may well not be mergeable right now. It's kept separate from the backend commits so it's trivial to drop, or leave as a reference to rebase onto the SSR admin later. The feature works fully over the API without it. Flagging it here so you can decide whether to take it.

image

### Testing

Ran a consent matrix over a double and a single opt-in list: a `city = 'Berlin'` segment reached only the eligible confirmed Berlin subscribers and correctly skipped unconfirmed-on-double-optin, unsubscribed, blocklisted, and out-of-segment. Preview count, `to_send` and final `sent` all matched. Also checked 403 without the permission, 400 on a disallowed table, empty segment behaving as a normal send, and an empty-result segment finishing at `to_send = 0`. Running in production on a 130k-subscriber instance atm.


knadh:
Whew, this is a big PR which introduces serious changes to the core and also UX. I'm not in a position to review/validate/test/performance-test this right now. Also, this introduces changes to the frontend which is now obsolete (please see: https://github.com/knadh/listmonk/blob/master/CONTRIBUTING.md)

I will come back to this after the v7 release. There are a few other big PRs also that have been pending for a while (saved queries, webhooks etc.). Thanks @CookedApps.

CookedApps:
No hurry, happy to revisit this after the v7 release @knadh 🙌

PR #3134, opened 2026-07-08
Add API endpoint to upsert subscriptions with attribs

### LINKED ISSUE #3133: API for upserting a subscription
I subscribe my users via API, setting some attribs. But the API is pretty thorny to use when the user has received an opt-in email but hasn't clicked it.

1. `POST` to `/api/subscribers`, which returns a 409.
2. `GET` `/api/subscribers?search=email@example.com` to get the subscription ID via email
3. `POST` to `/api/subscribers/optin`

Instead, I wish there were a `POST` `/api/subscribers/subscription` endpoint that would let me "upsert" ("optsert"?) a user in one request.

FWIW, `/api/public/subscription` already works exactly the way I want. The public API will automatically create the subscriber if it doesn't exist, resending the opt-in email if needed, with just one request, instead of three. (If I could, I'd just switch to the public API, but I need to set attribs, so I have to use the authenticated API.)

dfabulich:
Fixes #3133

### DIFF
--- a/cmd/handlers.go
+++ b/cmd/handlers.go
@@ -121,6 +121,7 @@ func initHTTPHandlers(e *echo.Echo, a *App) {
 		g.GET("/api/subscribers/:id/bounces", pm(hasID(a.GetSubscriberBounces), "bounces:get"))
 		g.DELETE("/api/subscribers/:id/bounces", pm(hasID(a.DeleteSubscriberBounces), "bounces:manage"))
 		g.POST("/api/subscribers", pm(a.CreateSubscriber, "subscribers:manage"))
+		g.POST("/api/subscribers/subscription", pm(a.SubscribeSubscriber, "subscribers:manage"))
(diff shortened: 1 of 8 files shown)


knadh:
Hi @dfabulich. I've only been able to give this a cursory look so far, but this should be `/api/subscribers/:id/subscription` and not a multi-subscriber endpoint.

dfabulich:
I think it makes sense to do some bikeshedding on how exactly to spell the endpoint URL, but I don't think it makes sense to have an `:id` parameter in the URL.

The idea is that the API client has an email address and N list IDs, but the client doesn't know whether the email address is already subscribed or not, and certainly doesn't already know the ID of the existing subscriber (if any).

I'm trying to mirror the shape of `/api/public/subscription`'s endpoint, but in an authenticated API endpoint, so I can set some attribs, and I'm specifically trying to avoid making multiple API requests.

PR #3137, opened 2026-07-09
Attach newly created lists to the creating user's list role

### LINKED ISSUE #2798: User cannot see their own created lists - new lists not automatically assigned to user's role
Hi team,
I've found a permissions issue in Listmonk v5.1.0 where users with list creation permissions cannot see the lists they create.

**Version:**
- Listmonk version: 5.1.0
- Deployment: Docker container
- Database: PostgreSQL

**Steps to Reproduce**
1. Create a user role with create list permissions (e.g., "Marketing Team")
2. Assign a user to this role
3. Assign a user to a list role
4. Grant the user permission to view/manage lists assigned to their list role 
5. User logs in and creates a new list
6. User navigates to Lists page

**Expected Behavior**
When a user creates a new list, it should be automatically assigned to the user's list role, making it immediately visible and manageable by that user.

**User workflow should be:**
1. User creates list ✅
2. List is automatically associated with user's list role ✅ (expected)
3. User can see and manage the list ✅ (expected)

**Actual Behavior**
When a user creates a new list, it is NOT automatically assigned to any list role. The user cannot see or manage the list they just created.

**Current broken workflow:**
1. User creates list ✅
2. List is created but not assigned to any list role ❌
3. User cannot see their own list ❌
4. Admin must manually assign the list to the user's list role ❌

**Impact**
This creates a poor user experience where:

- Users think list creation failed (they can't see it)
- Admins must manually assign every new list to the correct role
- Users cannot work independently even with proper permissions

**Proposed Solution**
When a user creates a list, the list should be automatically assigned to:

- Option A: All roles that the user belongs to
- Option B: The user's primary/default role
- Option C: Allow the user to select which role(s) during list creation

TowyTowy:
A non-super-admin user who creates a list (with the `lists:manage_all` permission) could not see or manage it afterwards. New lists were not attached to any list role, so the creator had no `list:get` grant on them and they stayed invisible on the Lists page until an admin manually assigned them.

On creation, attach the new list to the creator's list role (if they have one) with get and manage permissions, so it is immediately visible and manageable. Super admins and users with `lists:get_all` / `lists:manage_all` already see all lists and are unaffected.

Backend-only (respects the frontend freeze). Verified with `go build` + `go vet`; the repo has no Go test harness for this path.

Fixes #2798

### DIFF
--- a/cmd/lists.go
+++ b/cmd/lists.go
@@ -108,6 +108,17 @@ func (a *App) CreateList(c echo.Context) error {
 		return err
 	}
 
+	// Attach the new list to the creating user's list role (if any) with get and
+	// manage permissions so that they can immediately see and manage the list they
+	// created. Super admins and users with lists:get_all / lists:manage_all can
+	// already access all lists, so this is moot for them.
+	user := auth.GetUser(c)
+	if user.UserRole.ID != auth.SuperAdminRoleID && user.ListRoleID != nil {
+		if err := a.core.AddListPermission(*user.ListRoleID, out.ID, []string{auth.PermListGet, auth.PermListManage}); err != nil {
+			return err
+		}
+	}
+
 	return c.JSON(http.StatusOK, okResp{out})
 }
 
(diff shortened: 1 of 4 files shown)


knadh:
Auto-amending a role created by a super-admin is not the right way to do it. I don't have a clear answer on how to go about solving it, but please refer to the other discussion threads in issues that have talked about this in the past.

PR #3140, opened 2026-07-10
docs: add Zenith Hosting deploy option

odpay:
hey, i work on Zenith Hosting, a startup doing one-click hosting for open source apps. we added listmonk to our marketplace, so this PR adds a small "Deploy with Zenith" badge under Installation (links to https://zenith.hosting/host/listmonk).

we share a cut of revenue with the projects we host. happy to explain if you're curious: hello@zenith.hosting

PR #3142, opened 2026-07-13
feat: add Zenith Hosting to hosting options

odpay:
hey @knadh, thanks for the review on #3140. as requested, this moves it to a "Deploy with Zenith" image button in the hosting providers section on the site (docs/site/layouts/index.html), matching the other providers' style, instead of amending the readme.

it links to https://zenith.hosting/host/listmonk. Zenith Hosting is a startup doing one-click managed hosting for open source apps, and we share a cut of revenue with the projects we host. happy to explain: hello@zenith.hosting


knadh:
Please commit the SVG button to the repo like the other buttons.

odpay:
done, committed the SVG locally like the others (docs/site/static/static/images/host-zenith.svg) and pointed index.html at it. thanks @knadh!

odpay:
hey @knadh, gentle bump on this one. the change you asked for is in: the svg is committed to the repo at `docs/site/static/static/images/host-zenith.svg` like the other buttons, and `index.html` points at the local copy rather than an external url.

anything else you want adjusted before this can go in?

knadh:
Sorry, I lost track of this. Was just reviewing, but it looks like your image file is not an actual SVG, but a raster embedded in an SVG container. Was unable to preview it on GitHub. Previewed in Inkscape and could see that it's a raster (renders blurred). Could you please replace that with a true vector SVG?

odpay:
good catch on the symptom, and thanks for actually opening it in inkscape rather than just eyeballing it.

it isn't a raster though. there's no `<image>` element and no base64 payload in there, it's 10 `<path>` elements straight out of figma. what you were seeing is two figma drop-shadow filters built on `feGaussianBlur`, plus a couple of `mask-type:luminance` masks. inkscape rasterizes filter regions when it renders, so it genuinely looks blurred there, and github's svg sanitizer strips filters and inline styles, which is why the preview came up empty.

either way you shouldn't have to care about any of that in your repo, so i've flattened it. the shadows and masks are gone, it's now 7 plain paths in a clipped group, no filters, no masks, no inline styles. 33kb down to 26kb, previews fine on github, crisp in inkscape.

pushed to this branch, should be the only change.

PR #3144, opened 2026-07-13
Add campaign delivery window pacing

esokullu:
## Summary

This adds campaign-level delivery window pacing so a single campaign can be spread evenly over a configured time window without manually splitting a list into many smaller campaigns.

New campaign delivery controls are available in the campaign form:

- **Send immediately**: existing behavior.
- **Spread over...**: sends the campaign over a configured duration such as `24h`.

New API fields:

- `send_until`: absolute timestamp that marks the end of the delivery window.
- `send_window`: convenience duration such as `24h`; it is converted to `send_until` relative to `send_at`, or to the current time when `send_at` is empty.

Example:

```json
{
  send_at: 2026-07-14T10:00:00Z,
  send_window: 24h
}
```

The campaign remains one campaign in the UI/API. It does not create child campaigns or separate drip records.

## Why

listmonk already has global throughput and sliding-window rate limits, which are useful for server/provider limits. This patch targets a different use case: per-campaign pacing for sender reputation, warmup, and scheduled newsletter delivery.

Without this, users who want to send one campaign over 12-24 hours have to manually create list segments and schedule many separate campaigns. That fragments analytics and is easy to misconfigure.

## Implementation Notes

- Adds nullable `campaigns.send_until` with a migration.
- Uses a fork-specific migration version name (`v6.2.1-evalsignal.1`) instead of claiming an upstream future release number.
- Exposes `send_until` on campaign models and API responses.
- Adds `send_window` as an API-only helper field for create/update requests.
- Adds campaign form UI controls for immediate vs paced delivery.
- Updates campaign validation so `send_until` must be after `send_at` or now.
- Paces fetching subscribers in the campaign pipe without blocking the shared manager loop.
- Avoids advancing `last_subscriber_id` ahead of unsent work and keeps progress on one campaign row.
- Existing behavior is unchanged when `send_until` / `send_window` are not set.

## Testing

- `go test ./...`
- `yarn --cwd frontend lint`
- `yarn --cwd frontend build`

The frontend build completed with existing dependency/runtime warnings from Vite/Sass/Bulma, but no build failure.


knadh:
Please see this: https://github.com/knadh/listmonk/issues/3073

I'm not accepting changes to the Vue frontend until the v7 release.

Issue #2693, opened 2025-10-01
Support ongoing “Welcome” campaigns that stay active for new subscribers


Rush:
**Is your feature request related to a problem? Please describe.**
Currently, campaigns in Listmonk are treated as finite: once a campaign is sent, it is marked as complete and no longer applies to new subscribers. This makes it difficult to set up a simple, ongoing “Welcome” email flow where anyone who joins a list — even after the campaign has been launched — automatically receives the message.

**Describe the solution you’d like**
It would be helpful to introduce a new campaign type (e.g. ongoing campaign) or a checkbox option that, when enabled, keeps a campaign active. New subscribers joining the list would automatically receive the campaign email(s), even if they subscribe after the initial send.

**Workaround**
A current workaround is to call the Send test email API endpoint on an inactive or finished campaign each time a subscriber is added via API. However, this is not ideal and adds unnecessary complexity to what should be a straightforward automation use case.

candidexmedia:
> A current workaround is to call the Send test email API endpoint on an inactive or finished campaign each time a subscriber is added via API. 

I started playing around with [Zapier + Listmonk API](https://github.com/knadh/listmonk/discussions/2304), and I'm curious about how you got this working. My API setup knowledge pretty much ends with nocode tools like that one, so I'm curious about how to set that automation up.

k8ieone:
still relevant

Arturgrr:
Opened #3151 for this. It overlaps with #1818, but differs in a few ways:

- **Inline content per list** — subject + full editor (richtext/HTML/markdown/visual) + optional base template, rather than only picking a pre-made transactional template.
- **Guaranteed once-only** — a `subscriber_welcomes` table with an atomic claim sends the welcome at most once per (subscriber, list), including on re-subscribe and concurrent events (#1818 dedupes only within a single request).
- **Bulk import** — CSV import can welcome only the newly-added subscribers.

Same trigger semantics: single opt-in sends on subscribe, double opt-in after confirmation.

PR #3151, opened 2026-07-17
Add per-list welcome emails automatically sent to new subscribers

Arturgrr:
Implements #2693.

Adds an optional per-list welcome email, sent automatically when a subscriber becomes an active member of a list (single opt-in: on subscription; double opt-in: after confirmation). Delivery is deduplicated per (subscriber, list), so it is sent at most once.

- Configured per list via a new "Welcome email" tab (subject, format, template, and body with subscriber personalisation).
- Reuses the existing campaign render/send pipeline via a cached, non-persisted campaign per list — no new delivery path.
- CSV import gains an opt-in to send the welcome email to newly-added subscribers only.
- Schema changes ship in migration v6.3.0 (welcome_* columns on lists, subscriber_welcomes table).

Issue: https://github.com/knadh/listmonk/issues/2693


knadh:
Hi @Arturgrr. Please see: https://github.com/knadh/listmonk/issues/3073 and WIP https://github.com/knadh/listmonk/pull/2839 which will be the foundation for a feature like this.

PR #3161, opened 2026-07-23
Add support for Global sending calendar for campaigns

VikranthBala:
## Summary
This introduces a Global Sending Calendar feature to listmonk, allowing administrators to restrict campaign email delivery to specific days of the week and hours of the day.

New performance configuration settings are available in the UI:

Enable global sending calendar: Toggle to enable/disable the calendar check.
Weekly Schedule Calendar Rules: Custom day-of-week active hour windows (e.g. Monday 09:00 to 17:00).
When enabled, listmonk will check the calendar schedule before sending campaign batches and sleep until the next allowed window.

## Why
While listmonk supports global throughput and sliding-window rate limits to manage server/provider capacities, users often need to restrict email delivery to specific business hours or weekdays (e.g. to maximize open rates, avoid sending newsletters on weekends, or respect subscriber timezone behaviors).

This feature allows scheduling campaign delivery window constraints globally without requiring manual list fragmentation or scheduling separate campaigns.

## Implementation Notes
Top-level Calendar Check: Performs the SendCalendar check at the very beginning of the NextSubscribers function in 

* **pipe.go**
 (before querying the database). This prevents advancing the DB subscriber checkpoint (last_subscriber_id) if the campaign needs to sleep outside a window.
* **Interruptible Sleep & Early Exits**: Added checks for p.stopped.Load() at the start of NextSubscribers and inside the subscriber push loop. Sleeping is done using the interruptible p.wait(wait) helper so that pausing a campaign during a calendar wait window stops the campaign instantly.
* **Sliding Window Integration**: Integrates the calendar check inside the sliding window sleep block. If the end of a sliding window sleep falls outside the calendar schedule window, the wait duration is automatically extended to sleep until the next calendar window opens, preventing campaign deliveries from spilling outside permitted hours.

## Testing
Build Validation: Run go build ./... to verify that the changes compile cleanly.

### Manual Verification:
Configured a global sending calendar rule (e.g., active Monday through Friday 09:00 - 17:00). Starting a campaign outside these hours correctly sleeps.
Pausing during the calendar wait stops the pipe immediately. Resuming the campaign starts a new pipe, checks the calendar, and sleeps/resumes properly without skipping any subscriber checkpoints.
Verified that sliding window limit sleeps correctly calculate and sum up the wait time if the wake-up time crosses into a restricted calendar window.


knadh:
Please see this: https://github.com/knadh/listmonk/issues/3073

I'm not accepting changes to the Vue frontend until the v7 release.

PR #3208, opened 2026-08-30
security: limit HTTP request body size to prevent memory-exhaustion DoS

AARUSHGAUR:
## Summary
Several handlers read the request body (settings, media uploads, bounce and webhook endpoints) via `io.ReadAll` on `r.Body` without an explicit size limit. An attacker can send an unbounded request body and exhaust server memory (denial of service).

`codesentry` (static analysis) flagged 9 such handlers in `cmd/` and `internal/`.

## Fix
Register Echo's built-in `middleware.BodyLimit` (default `32M`) on the main HTTP server in `initHTTPServer`. It wraps each request body with `http.MaxBytesReader`, so the limit applies to all routes — including the flagged handlers — without touching them individually.

## Notes
- Default is `32M`; happy to make it a configurable `app.max_request_body_size` setting if you'd prefer it tunable rather than hard-coded.
- This is defense-in-depth hardening; behavior is unchanged for bodies under the limit.

Verified with `go build ./...`.

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -51,6 +51,7 @@ import (
 	"github.com/knadh/listmonk/models"
 	"github.com/knadh/stuffbin"
 	"github.com/labstack/echo/v4"
+	"github.com/labstack/echo/v4/middleware"
 	"github.com/lib/pq"
 	flag "github.com/spf13/pflag"
 	"gopkg.in/volatiletech/null.v6"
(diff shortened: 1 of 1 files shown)


AARUSHGAUR:
Hi! The GitHub Actions workflow for this PR is currently waiting for maintainer approval because this is a first-time fork PR.
Could you please approve workflow run 33289537539 from the Actions tab so the CI checks can run?
Thank you!

knadh:
Hardcoding 32 MB doesn't make sense. There are instances where files to the tune of 100s of MBs (even GBs) are uploaded via the Media UI to be linked in e-mails.

PR #3216, opened 2026-09-04
docs: the backup warnings name the uploads directory as well as the database

### LINKED ISSUE #3215: Docs: "take a backup of the Postgres database" leaves every uploaded media file behind
**Environment**

- listmonk v6.2.0, `listmonk/listmonk:v6.2.0` with `postgres:17-alpine`, deployed from
  the project's own `docker-compose.yml` with the image pinned.
- Stock configuration: media provider `filesystem`, upload path unchanged.

**What the documentation says**

There is no backup page. The word "backup" appears twice, both as warnings attached to
something else:

- upgrade page: "Always take a backup of the Postgres database before upgrading
  listmonk"
- installation page, on nightlies: "Always take a backup of your Postgres database
  before using a nightly release"

Both name the Postgres database and nothing else.

**What I did**

1. Started listmonk from the project's compose file.
2. Created a list and a subscriber, and uploaded one image through `/api/media`.
   With a stock configuration listmonk wrote two files:

   ```text
   /listmonk/uploads/drill-canary.png
   /listmonk/uploads/thumb_drill-canary.png
   ```

3. Took the documented backup: `pg_dump -U listmonk -d listmonk` (71 KiB).
4. Destroyed everything, restored the dump into a fresh Postgres, and started listmonk
   against it.

**What I observed**

Everything in the database came back - lists, subscribers, the admin account, settings,
and the row in `media` naming the uploaded file. The file itself is gone, so the media
library lists an image that cannot be loaded, and so does every campaign that used it.

**Why I think this is worth a documentation change**

The `media` table stores a file name; the file lives on disk. A person who follows the
documentation exactly has a backup that restores their database perfectly and loses
every image they have ever uploaded, with no error at any point - the restore looks
completely successful.

The project clearly knows the directory exists: `docker-compose.yml` mounts it. The
comment there says "To use this, change directory path in Admin -> Settings -> Media to
/listmonk/uploads", which reads as if the mount is optional - but the default upload
path already resolves to `/listmonk/uploads` inside the image, so files land there
without changing any setting.

**Suggested change**

A short backup section, or two lines wherever the existing warning appears:

> To back up listmonk, take a `pg_dump` of the Postgres database **and** a copy of the
> media uploads directory - `/listmonk/uploads` in the Docker image, or whatever
> Admin -> Settings -> Media names. The database stores a row per uploaded file; the
> files themselves are only on disk. (If you use the S3 provider, the files are in your
> bucket instead.)

It would also be worth putting that somewhere other than the upgrade page. People
upgrade far more often than they restore, and the backup they take before an upgrade is
the one they will still have when something else goes wrong.

I would be glad to send a docs PR for this if it would be welcome.

---

Everything above, with the commands and the unedited output, is at <https://github.com/spelingbee/drillback/tree/main/docs/drill/listmonk> - one leg of a restore drill across fifteen self-hosted applications, each one following its own backup documentation as written.

spelingbee:
The two places the docs say "take a backup of the Postgres database" now also name the media uploads directory. The `media` table stores a file name per upload; the file itself is only on disk (`/listmonk/uploads` in the Docker image, or the path in Admin -> Settings -> Media), so a dump-only backup restores a media library whose images cannot be loaded. Tested on v6.2.0 with the project's compose file. Two lines, no new page, as proposed in the issue.

Closes #3215.

### DIFF
--- a/docs/docs/content/installation.md
+++ b/docs/docs/content/installation.md
@@ -77,7 +77,7 @@ command: [sh, -c, "./listmonk --install --idempotent --yes --config /listmonk/co
 ## Nightly
 
 !!! Warning
-    Nightly releases are untested and may have bugs. Use at your own risk. Always take a backup of your Postgres database before using a nightly release.
+    Nightly releases are untested and may have bugs. Use at your own risk. Always take a backup of your Postgres database and of the media uploads directory (`/listmonk/uploads` in the Docker image) before using a nightly release.
 
 A nightly build is automatically published with the latest changes merged to the repository. If you want to access the latest changes without waiting for versioned releases, you can obtain the nightly builds and follow the same instructions above.
 
--- a/docs/docs/content/upgrade.md
+++ b/docs/docs/content/upgrade.md
@@ -1,7 +1,7 @@
 # Upgrade
 
 !!! Warning
-    Always take a backup of the Postgres database before upgrading listmonk
+    Always take a backup of the Postgres database before upgrading listmonk, and a copy of the media uploads directory (`/listmonk/uploads` in the Docker image, or whatever Admin -> Settings -> Media names). The database stores a row per uploaded file; the files themselves are only on disk. If you use the S3 provider, the files are in your bucket instead.
 
 ## Binary
 - Stop the running instance of listmonk.


knadh:
Hi @spelingbee. listmonk upgrades can make changes to the DB structure, which poses a risk. It'll never do that to media files on the disk. There's no risk to that when doing an upgrade. I don't think it needs to be mentioned in the docs.

Is there a particular reason why you want to add this?

spelingbee:
Fair point, and thank you for looking. The reason I sent it: those two warnings are the only place the docs mention backups at all, so in practice they are the backup guidance people find. I restored from exactly the backup the upgrade page describes (pg_dump only) and the media library came back with rows pointing at files that no longer existed. You are right that the upgrade itself never touches the files, so the upgrade page is the wrong place to say it. If you would take a two-line note somewhere else - installation, or a short Backup section under Maintenance - saying "the database plus the uploads directory", I will send that instead. If you would rather not have it at all, I will close this. Either way, listmonk itself restored cleanly.

